PR #2, opened 2018-09-12
update README.md

mrogers950:
@openshift/sig-security

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,96 @@
-# service-ca
-Coming soon...
+# OpenShift Service CA Operator
 
-Controller to mint and manage serving certificates for Kubernetes services.
+This operator runs the following OpenShift controllers:
+* **service-ca controller:**
+  * Issues a signed serving certificate/key pair to services annotated with 'service.alpha.openshift.io/serving-cert-secret-name' via a secret. [See the current OKD documentation for usage.](https://docs.okd.io/latest/dev_guide/secrets.html#service-serving-certificate-secrets)
 
+* **configmap-cabundle-injector controller:**
+  * Watches for configmaps annotated with 'service.alpha.openshift.io/inject-cabundle=true' and adds or updates a data item (key "cabundle.crt") containing the PEM-encoded CA signing bundle. Consumers of the configmap can then trust cabundle.crt in their TLS client configuration, allowing connections to services that utilize service-serving certificates.
 
-Current thinking of how this will be structured (this is subject to change).
- 1. Controller to create service serving cert/key pairs as today
- 2. Controller to keep apiservices up to date with a CA bundle based on an annotation on the apiservice.
- 3. Controller to maintain (but not create) configmaps with a CA bundle.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,11 +1,97 @@
...
-Coming soon...
+# OpenShift Service CA Operator
 
-Controller to mint and manage serving certificates for Kubernetes services.
+This operator runs the following OpenShift service-serving certificate controllers:
+* service-ca controller:
+  * Issues serving certificates based on annotations on services, and provides them via secrets. [See the current OKD documentation for usage.](https://docs.okd.io/latest/dev_guide/secrets.html#service-serving-certificate-secrets)
 
+* configmap-cabundle-injector controller:
+  * Watches for configmaps in any namespace annotated with 'service.alpha.openshift.io/inject-cabundle=true' and adds or updates a data item (key "cabundle.crt") containing the PEM-encoded CA signing bundle. Consumers of the configmap can then trust cabundle.crt in their TLS client configuration, allowing connections to services that utilize service-serving certificates.
```
ericavonb: > any namespace annotated with 'service.alpha.openshift.io/inject-cabundle=true'
I wish there was a way to control this operator's scope via RBAC instead of namespace annotations 🙁 . Maybe one day.

enj: How are we expecting this annotation to be added to the namespace in the first place?  Normal users cannot edit a namespace.

README.md
```diff
@@ -1,11 +1,97 @@
-# service-ca
-Coming soon...
+# OpenShift Service CA Operator
 
-Controller to mint and manage serving certificates for Kubernetes services.
+This operator runs the following OpenShift service-serving certificate controllers:
```
ericavonb: Can we get rid of `service-serving` terminology? I find just "runs the following OpenShift controllers:" easier to understand and read.

README.md
```diff
@@ -1,11 +1,97 @@
...
+* service-ca controller:
+  * Issues serving certificates based on annotations on services, and provides them via secrets. [See the current OKD documentation for usage.](https://docs.okd.io/latest/dev_guide/secrets.html#service-serving-certificate-secrets)
```
ericavonb: How about `Issues a signed serving certificate/key pair to services annotated with <annotation here> via a secret.`? Something more specific about which assets are generated.

README.md
```diff
@@ -1,11 +1,97 @@
-# service-ca
-Coming soon...
+# OpenShift Service CA Operator
 
-Controller to mint and manage serving certificates for Kubernetes services.
+This operator runs the following OpenShift service-serving certificate controllers:
+* service-ca controller:
```
ericavonb: Make the controller names stand out more, like make bold or something

README.md
ericavonb: I think that's the point. It's a hack to give administrators control over who gets access to the CA in a dynamic way.

enj: But why would we ever need to limit who can get the CA?

mrogers950: I mean to say *configmaps* that are annotated, not namespaces, so it is badly worded

mrogers950:
@ericavonb updated

PR #5, opened 2018-09-13
set .travis.yml golang to 1.9

mrogers950:
1.10 didn't seem to work. This matches the original repo .travis.yml


eparis:
isn't that a problem, since openshift is going to built with 1.10.3 (or newer?)

PR #8, opened 2018-09-21
update to flat updatepayload structure

juanvallejo:
Updates Dockerfile to include a /manifests directory in image.

cc @deads2k @mfojtik @sttts @mrogers950


juanvallejo:
@mrogers950 your `./hack/update-generated-bindata.sh` script appears to be broken. It fails on [this line](https://github.com/openshift/service-ca-operator/blob/master/hack/update-generated-bindata.sh#L27) since it cannot find the `test/extended/...` path

mrogers950:
@juanvallejo we can close this for https://github.com/openshift/service-ca-operator/pull/9 , which updates the manifest files based on what we know now :)

PR #10, opened 2018-09-27
Add initial e2e testing

mrogers950:
Adds rudimentary tests. After merging I'll add e2e runs to the openshift/release CI configs, and then follow up with more tests.
/cc @openshift/sig-security

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -1,0 +1,171 @@
+package e2e
+
+import (
+	"math/rand"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/openshift/service-ca-operator/pkg/controller/servingcert"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -0,0 +1,179 @@
...
+	if !hasPodWithPrefixName(adminClient, caControllerPodPrefix, serviceCAControllerNamespace) {
+		t.Fatalf("%s not running in %s namespace", caControllerPodPrefix, serviceCAControllerNamespace)
+	}
+
+	// test the main feature. annotate service -> created secret
+	t.Run("serving-cert-annotation", func(t *testing.T) {
+		ns := createTestNamespace(adminClient)
+		if len(ns) == 0 {
+			t.Fatalf("could not create test namespace")
+		}
```
ericavonb: Should we clean up the namespace as well?

mrogers950: we do in cleanupServiceSignerTestObjects()

ericavonb: What about in the situation that `createServinCertAnnotatedService` fails below?

mrogers950: Good catch, fixed that.

mrogers950:
@ericavonb updated.

smarterclayton:
/test e2e-aws

(this is the gating job for your operator to ensure you don't break the install or other components)

I would recommend that you create at least 1 e2e test inside of the origin suite (the conformance suite) that verifies that service service certs are working correctly, if you haven't already.  that's your smoke test to ensure other people don't break you.  you are welcome to put additional tests there, since that is where tests that validate an openshift cluster is providing the required function go.  if you'd like to do more stress testing of your operator, that belongs here (in this repo).

PR #18, opened 2019-02-22
Make the signing CA last 1 year

stlaz:
Prolong the life of the service signing CA from 10 days to 730.


enj:
/lgtm

@openshift/sig-auth this is our commitment to make rotation work in 4.x where x is small.  Hope everyone is comfortable with that.

PR #17, opened 2019-02-22
Set service CA's management state to Managed

stlaz:
SSCS now goes to no-op mode if service CA operator is running. Take
over its duties.

PR #22, opened 2019-02-26
fix(**clusterrole.yaml): add pod privileges to controllers

ericavonb:
from #17 
```
W0226 00:28:08.608116       1 cmd.go:134] Using insecure, self-signed certificates
I0226 00:28:08.608310       1 crypto.go:493] Generating new CA for openshift-service-serving-cert-signer-apiservice-injector-signer@1551140888 cert, and key in /tmp/serving-cert-151717941/serving-signer.crt, /tmp/serving-cert-151717941/serving-signer.key
I0226 00:28:09.194355       1 observer_polling.go:40] Starting from specified content for file "/var/run/configmaps/config/controller-config.yaml"
I0226 00:28:09.194415       1 observer_polling.go:46] Adding reactor for file "/var/run/secrets/serving-cert/tls.crt"
I0226 00:28:09.194435       1 observer_polling.go:46] Adding reactor for file "/var/run/secrets/serving-cert/tls.key"
I0226 00:28:09.195036       1 observer_polling.go:106] Starting file observer
panic: unable to obtain replicaset reference for events: pods is forbidden: User "system:serviceaccount:openshift-service-cert-signer:apiservice-cabundle-injector-sa" cannot list resource "pods" in API group "" in the namespace "openshift-service-cert-signer"

goroutine 1 [running]:
github.com/openshift/service-ca-operator/vendor/github.com/openshift/library-go/pkg/controller/controllercmd.(*ControllerBuilder).Run(0xc4201da000, 0xc42041a110, 0x1d7ffc0, 0xc42027c800, 0x3, 0x3)
	/go/src/github.com/openshift/service-ca-operator/vendor/github.com/openshift/library-go/pkg/controller/controllercmd/builder.go:172 +0xc14
github.com/openshift/service-ca-operator/vendor/github.com/openshift/library-go/pkg/controller/controllercmd.(*ControllerCommandConfig).StartController(0xc420493600, 0x1d80000, 0xc420052038, 0x2b92f70, 0x0)
	/go/src/github.com/openshift/service-ca-operator/vendor/github.com/openshift/library-go/pkg/controller/controllercmd/cmd.go:193 +0x638
github.com/openshift/service-ca-operator/vendor/github.com/openshift/library-go/pkg/controller/controllercmd.(*ControllerCommandConfig).NewCommand.func1(0xc4207daf00, 0xc42052eae0, 0x0, 0x2)
	/go/src/github.com/openshift/service-ca-operator/vendor/github.com/openshift/library-go/pkg/controller/controllercmd/cmd.go:68 +0x271
github.com/openshift/service-ca-operator/vendor/github.com/spf13/cobra.(*Command).execute(0xc4207daf00, 0xc42052ea60, 0x2, 0x2, 0xc4207daf00, 0xc42052ea60)
	/go/src/github.com/openshift/service-ca-operator/vendor/github.com/spf13/cobra/command.go:760 +0x2c1
github.com/openshift/service-ca-operator/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc4201dca00, 0xc4207daf00, 0xc4207dac80, 0xc4201dd180)
	/go/src/github.com/openshift/service-ca-operator/vendor/github.com/spf13/cobra/command.go:846 +0x30a
github.com/openshift/service-ca-operator/vendor/github.com/spf13/cobra.(*Command).Execute(0xc4201dca00, 0x1c8e310, 0x2b71980)
	/go/src/github.com/openshift/service-ca-operator/vendor/github.com/spf13/cobra/command.go:794 +0x2b
main.main()
	/go/src/github.com/openshift/service-ca-operator/cmd/service-ca-operator/main.go:32 +0x187
```

Event recorder needs to be able to trace pods and such as related objects.


bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
```diff
@@ -29,3 +29,21 @@ rules:
   - events
   verbs:
   - create
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - secrets
```
enj: I do not think this controller needs read access to all secrets in all namespaces.  It might need some, but not all.

ericavonb: I was trying to get it to work to see which perms were needed, but since CI wasn't working: 6c4bf1a5c1dd2c46a0121883204997d2ec633e42

ericavonb:
It's green! @enj what do you think of the rbac changes here?

bindata/v4.0.0/configmap-cabundle-controller/clusterrole.yaml
```diff
@@ -12,19 +12,4 @@ rules:
...
-  - list
-  - watch
-  - update
-  - create
-- apiGroups:
-  - ""
-  resources:
-  - events
-  verbs:
   - create
```
enj: should not need create

bindata/v4.0.0/configmap-cabundle-controller/role.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
enj: I believe these are all the same and cover the namespace scoped library-go functionality.  We should probably just share a single role here.  A single role binding with all three controllers would also make sense.

pkg/operator/sync_apiservice_v4_00.go
```diff
@@ -44,6 +44,18 @@ func syncAPIServiceController_v4_00_to_latest(c serviceCAOperator, operatorConfi
 		return fmt.Errorf("%q: %v", "clusterrolebinding", err)
 	}
 
+	requiredRole := resourceread.ReadRoleV1OrDie(v4_00_assets.MustAsset("v4.0.0/apiservice-cabundle-controller/role.yaml"))
```
enj: should be able to do this once instead of three times with the shared role and binding

bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
```diff
@@ -23,9 +23,3 @@ rules:
   - watch
   - update
   - create
```
enj: I know this is preexisting, but I do not think this needs access to config maps at all.

bindata/v4.0.0/service-serving-cert-signer-controller/clusterrole.yaml
```diff
@@ -36,9 +36,3 @@ rules:
   - watch
   - update
   - create
```
enj: this also should not need config map access.

also not sure if it needs delete on secrets.

bindata/v4.0.0/configmap-cabundle-controller/role.yaml
ericavonb: I was weighing the trade off between less moving parts with sharing a role/rolebinding, and maintaining the independence of each controller. Like, right now you can delete any of the controllers and it's respective resources and the other run just fine; or you can run them independently with its set of resources. WDYT?

bindata/v4.0.0/service-serving-cert-signer-controller/clusterrole.yaml
ericavonb: I don't think it needs create on services either: a84d0954

bindata/v4.0.0/configmap-cabundle-controller/clusterrole.yaml
ericavonb: 👍 I'm all for removing what permissions we can. a84d0954

bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
ericavonb: a84d0954

bindata/v4.0.0/apiservice-cabundle-controller/defaultconfig.yaml
```diff
@@ -1,3 +1,7 @@
 apiVersion: servicecertsigner.config.openshift.io/v1alpha1
 kind: APIServiceCABundleInjectorConfig
 caBundleFile: /var/run/configmaps/signing-cabundle/cabundle.crt
+authentication:
+  disabled: true
+authorization:
+  disabled: true
```
ericavonb: :( fill fix

ericavonb:
@enj I'd rather leave the consolidating roles and rolebindings for later. The configmap access was needed for leader election, so had to add it back.

bindata/v4.0.0/configmap-cabundle-controller/role.yaml
mrogers950: I think the controllers can be coupled together if that makes deployments simpler, there's no situation where you would want to only run a subset of them.

enj:
/lgtm

We can sort out RBAC / bindata over time.

PR #23, opened 2019-02-26
OpenAPIV3 validation for operator CRD

sallyom:
@mrogers950  I used [this generator](https://github.com/openshift/kubernetes-sigs-controller-tools/cmd/crd) to generate the CRD.

A few tweaks after generation:
- add `singular: serviceca` bc this comment doesn't exist in openshift/api/operator/v1/types_serviceca.go: `// +serviceca:singular=serviceca`
- removed generated status fields, don't need to validate those 
-  removed `required` for most fields

this PR is similar to [this](https://github.com/openshift/cluster-authentication-operator/pull/80)

### DIFF
--- a/manifests/0000_09_service-ca-operator_02_crd.yaml
+++ b/manifests/0000_09_service-ca-operator_02_crd.yaml
@@ -3,12 +3,99 @@ kind: CustomResourceDefinition
 metadata:
   name: servicecas.operator.openshift.io
 spec:
-  scope: Cluster
   group: operator.openshift.io
-  version: v1
   names:
     kind: ServiceCA
     plural: servicecas
     singular: serviceca
-  subresources:
-    status: {}
+  scope: Cluster
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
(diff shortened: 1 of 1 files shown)


manifests/0000_09_service-ca-operator_02_crd.yaml
```diff
@@ -3,12 +3,116 @@ kind: CustomResourceDefinition
...
+                - 'Normal'
+                - 'Debug'
+                - 'Trace'
+                - 'TraceAll'
+            managementState:
+              description: managementState indicates whether and how the operator
+                should manage the component
+              type: string
+              enum:
+                - 'Force'
```
mrogers950: The operator currently sets managementState to "Paused", and uses that as a no-op during the sync. For the time being we should probably allow a "Paused" value

manifests/0000_09_service-ca-operator_02_crd.yaml
```diff
@@ -3,12 +3,116 @@ kind: CustomResourceDefinition
...
+        spec:
+          properties:
+            logLevel:
+              description: logLevel is an intent based logging for an overall component.  It
+                does not give fine grained control, but it is a simple way to manage
+                coarse grained logging choices that operators have to interpret for
+                their operands.
+              type: string
+              enum:
+                - 'Normal'
```
mrogers950: The default operator config sets logLevel to "4", and the controller code expects that. We should probably allow that value for the time being.

sallyom: I see, I'm going to remove the `enum`s then, I was under the impression the logLevel would _always_ be 1 of the 4 listed in api/operator/v1/types.go, but since this isn't the case, we should leave out the enums altogether, I think.

manifests/0000_09_service-ca-operator_02_crd.yaml
sallyom: I've removed the enums, see note below.

ericavonb:
/lgtm

I'm in favor of adding stricter validation but as it's friday afternoon, getting some validation into the release now and opening further prs on monday seems like the best way forward.

PR #24, opened 2019-02-27
Honor beta SSCS annotations

mrogers950:
This PR makes it so that `service.beta.openshift.io` variants of the annotations are honored, and both alpha and beta variants of the accounting annotations (errors, expiry, etc) are added to objects to make it easier to transition to beta completely. 
/cc @openshift/sig-auth

### DIFF
--- a/pkg/controller/api/api.go
+++ b/pkg/controller/api/api.go
@@ -7,14 +7,14 @@ import (
 )
 
 const (
-	InjectCABundleAnnotationName = "service.alpha.openshift.io/inject-cabundle"
-	BetaInjectCABundleAnnotationName = "service.beta.openshift.io/inject-cabundle"
-	InjectionDataKey             = "service-ca.crt"
+	InjectCABundleAnnotationName      = "service.beta.openshift.io/inject-cabundle"
+	AlphaInjectCABundleAnnotationName = "service.alpha.openshift.io/inject-cabundle"
+	InjectionDataKey                  = "service-ca.crt"
 )
 
 func HasInjectCABundleAnnotation(metadata v1.Object) bool {
-	return strings.EqualFold(metadata.GetAnnotations()[InjectCABundleAnnotationName], "true") ||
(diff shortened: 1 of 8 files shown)


pkg/controller/servingcert/controller/secret_creating_controller_test.go
```diff
@@ -169,6 +176,99 @@ func TestBasicControllerFlow(t *testing.T) {
...
+
+	if !foundSecret {
+		t.Errorf("secret wasn't created.  Got %v\n", kubeclient.Actions())
+	}
+	if !foundServiceUpdate {
+		t.Errorf("service wasn't updated.  Got %v\n", kubeclient.Actions())
+	}
+}
+
+func TestBasicControllerFlowBetaAnnotation(t *testing.T) {
```
stlaz: Could the gist of these two functions be unified in just one and be run from two different wrapping ones?

pkg/controller/api/api.go
```diff
@@ -8,11 +8,13 @@ import (
 
 const (
 	InjectCABundleAnnotationName = "service.alpha.openshift.io/inject-cabundle"
+	BetaInjectCABundleAnnotationName = "service.beta.openshift.io/inject-cabundle"
```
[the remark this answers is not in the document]
mrogers950: That's a good suggestion, fixed.

pkg/controller/servingcert/controller/secret_creating_controller_test.go
mrogers950: I think this would need a refactor for each of these specific variant functions, and we will only support the dual annotations for a period of time. I'd like to follow up with that at a later time if we need to. I did reduce the copy-pasta on the other tests, though.

stlaz: Lets' keep it that way, then, no reason to do any major code shoveling to get this PR in :wink:

stlaz:
/lgtm
Since we turned the state to Managed on Friday, I re-run all the tests, if they pass this should be safe to be merged.

PR #26, opened 2019-03-04
Bug 1683058: Prevent sharing of configmap and secret data

sallyom:
This PR will ensure changes in ca bundle injector configmap and serving cert secret data are stomped on by the controller.

https://jira.coreos.com/browse/AUTH-204

### DIFF
--- a/pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
+++ b/pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
@@ -46,20 +46,21 @@ func (ic *configMapCABundleInjectionController) Sync(obj v1.Object) error {
 	if !api.HasInjectCABundleAnnotation(sharedConfigMap) {
 		return nil
 	}
+	// ensure data of configmap
+	return ic.ensureConfigMapCABundleInjection(sharedConfigMap)
+}
+
+// ensureConfigMapCABundleInjection will create or update configmap for the
+// CA bundle injection as appropriate.
+func (ic *configMapCABundleInjectionController) ensureConfigMapCABundleInjection(current *corev1.ConfigMap) error {
 	// skip updating when the CA bundle is already there
-	if data, ok := sharedConfigMap.Data[api.InjectionDataKey]; ok && data == ic.ca {
+	if data, ok := current.Data[api.InjectionDataKey]; ok &&
+		data == ic.ca && len(current.Data) == 1 {
 		return nil
 	}
-
 	// make a copy to avoid mutating cache state
-	configMapCopy := sharedConfigMap.DeepCopy()
-
-	if configMapCopy.Data == nil {
-		configMapCopy.Data = map[string]string{}
-	}
-
-	configMapCopy.Data[api.InjectionDataKey] = ic.ca
-
-	_, err := ic.configMapClient.ConfigMaps(configMapCopy.Namespace).Update(configMapCopy)
+	configMapCopy := current.DeepCopy()
+	configMapCopy.Data = map[string]string{api.InjectionDataKey: ic.ca}
+	_, err := ic.configMapClient.ConfigMaps(current.Namespace).Update(configMapCopy)
 	return err
 }
--- a/pkg/controller/configmapcainjector/controller/configmap_injecting_controller_test.go
+++ b/pkg/controller/configmapcainjector/controller/configmap_injecting_controller_test.go
@@ -31,6 +31,9 @@ func validateActions(t *testing.T, expectedActionsNum int, actions []clienttesti
 	if expected := "content"; string(actual.Data[api.InjectionDataKey]) != expected {
 		t.Error(diff.ObjectDiff(expected, actual))
 	}
+	if len(actual.Data) != 1 {
+		t.Errorf("expected only one data key in CM, got %#v", actual.Data)
+	}
 }
 
 func TestSyncConfigMapCABundle(t *testing.T) {
@@ -77,6 +80,26 @@ func TestSyncConfigMapCABundle(t *testing.T) {
 					},
 					Data: map[string]string{
 						api.InjectionDataKey: "foo",
+					},
+				},
+			},
+			namespace:          "foo",
+			cmName:             "foo",
+			caBundle:           "content",
+			expectedActionsNum: 1,
+		},
+		{
+			name: "requested and same and added-data",
+			startingConfigMaps: []runtime.Object{
+				&v1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:        "foo",
+						Annotations: map[string]string{api.InjectCABundleAnnotationName: "true"},
+						Namespace:   "foo",
+					},
+					Data: map[string]string{
+						api.InjectionDataKey: "content",
+						"blah":               "foo",
 					},
 				},
 			},
--- a/pkg/controller/servingcert/controller/secret_creating_controller.go
+++ b/pkg/controller/servingcert/controller/secret_creating_controller.go
@@ -288,8 +288,7 @@ func toBaseSecret(service *v1.Service) *v1.Secret {
 		Type: v1.SecretTypeTLS,
 	}
 }
-
-func toRequiredSecret(dnsSuffix string, ca *crypto.CA, service *v1.Service, secretCopy *v1.Secret) error {
+func getServingCert(dnsSuffix string, ca *crypto.CA, service *v1.Service) (*crypto.TLSCertificateConfig, error) {
 	dnsName := service.Name + "." + service.Namespace + ".svc"
 	fqDNSName := dnsName + "." + dnsSuffix
 	certificateLifetime := 365 * 2 // 2 years
@@ -299,24 +298,31 @@ func toRequiredSecret(dnsSuffix string, ca *crypto.CA, service *v1.Service, secr
 		cryptoextensions.ServiceServerCertificateExtensionV1(service),
 	)
 	if err != nil {
+		return nil, err
+	}
+	return servingCert, nil
+}
+
+func toRequiredSecret(dnsSuffix string, ca *crypto.CA, service *v1.Service, secretCopy *v1.Secret) error {
+	servingCert, err := getServingCert(dnsSuffix, ca, service)
+	if err != nil {
 		return err
 	}
 	certBytes, keyBytes, err := servingCert.GetPEMBytes()
 	if err != nil {
 		return err
 	}
-
 	if secretCopy.Annotations == nil {
 		secretCopy.Annotations = map[string]string{}
 	}
-	if secretCopy.Data == nil {
-		secretCopy.Data = map[string][]byte{}
+	// let garbage collector cleanup map allocation, for simplicity
+	secretCopy.Data = map[string][]byte{
+		v1.TLSCertKey:       certBytes,
+		v1.TLSPrivateKeyKey: keyBytes,
 	}
 
 	secretCopy.Annotations[api.AlphaServingCertExpiryAnnotation] = servingCert.Certs[0].NotAfter.Format(time.RFC3339)
 	secretCopy.Annotations[api.ServingCertExpiryAnnotation] = servingCert.Certs[0].NotAfter.Format(time.RFC3339)
-	secretCopy.Data[v1.TLSCertKey] = certBytes
-	secretCopy.Data[v1.TLSPrivateKeyKey] = keyBytes
 
 	ocontroller.EnsureOwnerRef(secretCopy, ownerRef(service))
 
--- a/pkg/controller/servingcert/controller/secret_updating_controller.go
+++ b/pkg/controller/servingcert/controller/secret_updating_controller.go
@@ -1,9 +1,12 @@
 package controller
 
 import (
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"time"
 
+	"github.com/golang/glog"
 	"k8s.io/api/core/v1"
 	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -69,70 +72,140 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
 
 func (sc *serviceServingCertUpdateController) Sync(obj metav1.Object) error {
 	sharedSecret := obj.(*v1.Secret)
+	service := sc.getServiceForSecret(sharedSecret)
+	if service == nil {
+		return nil
+	}
 
-	regenerate, service := sc.requiresRegeneration(sharedSecret)
-	if !regenerate {
+	if !isSecretValidForService(service, sharedSecret) {
 		return nil
 	}
 
 	// make a copy to avoid mutating cache state
 	secretCopy := sharedSecret.DeepCopy()
 
-	if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
+	if sc.requiresRegeneration(service, sharedSecret, sc.minTimeLeftForCert) {
+		// Regenerate the secret
+		if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
+			return err
+		}
+		_, err := sc.secretClient.Secrets(secretCopy.Namespace).Update(secretCopy)
 		return err
 	}
-
-	_, err := sc.secretClient.Secrets(secretCopy.Namespace).Update(secretCopy)
-	return err
+	// If not regenerating, perform checks here to
+	// 1. ensure only the 2 data keys we want exist, TLSCertKey and TLSPrivateKeyKey
+	// 2. ensure cert data is at least a parseable certificate, if not replace with valid data
+	// This does not ensure that somebody didn't swap out secret data for another valid cert.
+	update, err := sc.ensureSecretData(service, secretCopy)
+	if err != nil {
+		return err
+	}
+	if update {
+		_, err := sc.secretClient.Secrets(secretCopy.Namespace).Update(secretCopy)
+		return err
+	}
+	return nil
 }
 
-func (sc *serviceServingCertUpdateController) requiresRegeneration(secret *v1.Secret) (bool, *v1.Service) {
-	serviceName, ok := toServiceName(secret)
-	if !ok {
-		return false, nil
-	}
-
-	sharedService, err := sc.serviceLister.Services(secret.Namespace).Get(serviceName)
-	if kapierrors.IsNotFound(err) {
-		return false, nil
-	}
-	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("unable to get service %s/%s: %v", secret.Namespace, serviceName, err))
-		return false, nil
-	}
-
+func isSecretValidForService(sharedService *v1.Service, secret *v1.Secret) bool {
+	isValid := true
 	if sharedService.Annotations[api.ServingCertSecretAnnotation] != secret.Name && sharedService.Annotations[api.AlphaServingCertSecretAnnotation] != secret.Name {
-		return false, nil
+		isValid = false
 	}
 	if secret.Annotations[api.ServiceUIDAnnotation] != string(sharedService.UID) && secret.Annotations[api.AlphaServiceUIDAnnotation] != string(sharedService.UID) {
-		return false, nil
+		isValid = false
 	}
+	return isValid
+}
 
+func (sc *serviceServingCertUpdateController) getServiceForSecret(sharedSecret *v1.Secret) *v1.Service {
+	serviceName, ok := toServiceName(sharedSecret)
+	if !ok {
+		return nil
+	}
+	service, err := sc.serviceLister.Services(sharedSecret.Namespace).Get(serviceName)
+	if kapierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("unable to get service %s/%s: %v", sharedSecret.Namespace, serviceName, err))
+		return nil
+	}
+	return service
+}
+
+func (sc *serviceServingCertUpdateController) requiresRegeneration(service *v1.Service, secret *v1.Secret, minTimeLeft time.Duration) bool {
 	// if we don't have an ownerref, just go ahead and regenerate.  It's easier than writing a
 	// secondary logic flow.
-	if !ocontroller.HasOwnerRef(secret, ownerRef(sharedService)) {
-		return true, sharedService
+	if !ocontroller.HasOwnerRef(secret, ownerRef(service)) {
+		return true
 	}
-
 	// if we don't have the annotation for expiry, just go ahead and regenerate.  It's easier than writing a
 	// secondary logic flow that creates the expiry dates
 	expiryString, ok := secret.Annotations[api.ServingCertExpiryAnnotation]
 	if !ok {
 		expiryString, ok = secret.Annotations[api.AlphaServingCertExpiryAnnotation]
 		if !ok {
-			return true, sharedService
+			return true
 		}
 	}
 	expiry, err := time.Parse(time.RFC3339, expiryString)
 	if err != nil {
-		return true, sharedService
+		return true
 	}
 
 	if time.Now().Add(sc.minTimeLeftForCert).After(expiry) {
-		return true, sharedService
+		return true
 	}
 
-	return false, nil
+	return false
+}
+
+func (sc *serviceServingCertUpdateController) ensureSecretData(service *v1.Service, secretCopy *v1.Secret) (bool, error) {
+	update := false
+	tlsCert, ok := secretCopy.Data[v1.TLSCertKey]
+	tlsKey, ok2 := secretCopy.Data[v1.TLSPrivateKeyKey]
+	if ok && ok2 {
+		// Don't update, replace unless len data is not 2
+		// This takes care of case where someone added an extra data field. Remove all but 2 keys
+		if len(secretCopy.Data) != 2 {
+			// don't regenerate, just get rid of bad data
+			secretCopy.Data = map[string][]byte{
+				v1.TLSCertKey:       tlsCert,
+				v1.TLSPrivateKeyKey: tlsKey,
+			}
+			update = true
+		}
+	} else {
+		// if required tlscertkey,tlsprivatekey fields missing, replace with valid secret
+		// Regenerate the secret
+		if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
+			return update, err
+		}
+		return true, nil
+	}
+	// This ensures someone did not replace secret cert data with ascii art
+	// Check for valid cert data, replace with valid data if not
+	block, _ := pem.Decode([]byte(tlsCert))
+	if block == nil {
+		// Regenerate the secret
+		glog.Infof("Error decoding cert bytes %s from secret: %s namespace: %s, replacing cert", v1.TLSCertKey, secretCopy.Name, secretCopy.Namespace)
+		// Regenerate the secret
+		if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
+			return update, err
+		}
+		return true, nil
+	}
+	_, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		glog.Infof("Error parsing %s from secret: %s namespace: %s, replacing cert", v1.TLSCertKey, secretCopy.Name, secretCopy.Namespace)
+		// Regenerate the secret
+		if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
+			return update, err
+		}
+		return true, nil
+	}
+	return update, nil
 }
 
 func toServiceName(secret *v1.Secret) (string, bool) {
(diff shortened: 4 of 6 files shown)


pkg/controller/servingcert/controller/secret_updating_controller_test.go
```diff
@@ -368,6 +368,29 @@ func TestRequiresRegenerationServiceUIDMismatchBetaAnnotation(t *testing.T) {
 			},
 			expected: true,
 		},
+		{
+			name: "secret with added data",
```
mrogers950: I don't think this tests your related changes as it is just exercising the check against an already-queued secret to see if it matches up with its owner (yours is good, so returning "no regeneration" is to be expected). You should make a test in the e2e suite that creates a serving cert like normal and then modifies the secret, and then checks to see if it gets stomped on or not.

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -58,9 +58,16 @@ func (sc *serviceServingCertUpdateController) addSecret(obj metav1.Object) bool
 }
 
 func (sc *serviceServingCertUpdateController) updateSecret(old, cur metav1.Object) bool {
-	// if the current doesn't have a service name, check the old
+	curSecret := cur.(*v1.Secret)
+	oldSecret := old.(*v1.Secret)
+	// if the current data does not match or doesn't have a service name, check the old
 	// TODO drop this
-	return sc.addSecret(cur) || sc.addSecret(old)
+	if string(curSecret.Data[v1.TLSCertKey]) == string(oldSecret.Data[v1.TLSCertKey]) &&
```
mrogers950: This seems right to me but my brain wants to think about the inverse case (i.e., cur and old data do not match or cur does not have 2 items, so add old). Not sure if that is clearer to anyone else :)

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -58,9 +58,16 @@ func (sc *serviceServingCertUpdateController) addSecret(obj metav1.Object) bool
...
+		string(curSecret.Data[v1.TLSPrivateKeyKey]) == string(oldSecret.Data[v1.TLSPrivateKeyKey]) &&
```
stlaz: `bytes.Equal()`

pkg/controller/servingcert/controller/secret_updating_controller.go
stlaz: +1, can't explain why my brain wants it that way.

test/e2e/e2e_test.go
```diff
@@ -19,9 +20,10 @@ import (
 )
 
 const (
-	serviceCAOperatorNamespace   = operatorclient.OperatorNamespace
-	serviceCAOperatorPodPrefix   = operatorclient.OperatorNamespace // Same as operator namespace
-	serviceCAControllerNamespace = operatorclient.TargetNamespace
+	serviceCAOperatorNamespace = operatorclient.OperatorNamespace
+	serviceCAOperatorPodPrefix = operatorclient.OperatorNamespace // Same as operator namespace
+	//serviceCAControllerNamespace = operatorclient.TargetNamespace
```
stlaz: Don't forget to remove this change :slightly_smiling_face:

test/e2e/e2e_test.go
```diff
@@ -74,6 +76,19 @@ func createServingCertAnnotatedService(client *kubernetes.Clientset, secretName,
 	return err
 }
 
+func createAnnotatedCABundleInjectionConfigMap(client *kubernetes.Clientset, configMapName, namespace string) error {
+	_, err := client.CoreV1().ConfigMaps(namespace).Create(&v1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: configMapName,
+			Annotations: map[string]string{
+				api.AlphaInjectCABundleAnnotationName: "true",
```
stlaz: Shouldn't we be testing this with the current version now? Or maybe with both?

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: lol ok, 2 against 1 :)

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: done, thanks

pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
```diff
@@ -46,20 +46,31 @@ func (ic *configMapCABundleInjectionController) Sync(obj v1.Object) error {
...
+func (ic *configMapCABundleInjectionController) ensureConfigMapCABundleInjection(current *corev1.ConfigMap) error {
 	// make a copy to avoid mutating cache state
-	configMapCopy := sharedConfigMap.DeepCopy()
-
+	configMapCopy := current.DeepCopy()
+	// skip updating when the CA bundle is already there
+	if data, ok := configMapCopy.Data[api.InjectionDataKey]; ok && data == ic.ca && len(configMapCopy.Data) == 1 {
+		return nil
+	}
 	if configMapCopy.Data == nil {
```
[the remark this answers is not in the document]
sallyom: good point. thanks :) done

pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
```diff
@@ -46,20 +46,31 @@ func (ic *configMapCABundleInjectionController) Sync(obj v1.Object) error {
 	if !api.HasInjectCABundleAnnotation(sharedConfigMap) {
 		return nil
 	}
-	// skip updating when the CA bundle is already there
-	if data, ok := sharedConfigMap.Data[api.InjectionDataKey]; ok && data == ic.ca {
-		return nil
+	// ensure data of configmap
+	if err := ic.ensureConfigMapCABundleInjection(sharedConfigMap); err != nil {
```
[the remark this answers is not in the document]
sallyom: good cleanup, thanks

test/e2e/e2e_test.go
sallyom: yup, thanks, done

pkg/controller/configmapcainjector/controller/configmap_injecting_controller_test.go
```diff
@@ -31,6 +31,9 @@ func validateActions(t *testing.T, expectedActionsNum int, actions []clienttesti
 	if expected := "content"; string(actual.Data[api.InjectionDataKey]) != expected {
 		t.Error(diff.ObjectDiff(expected, actual))
 	}
+	if expectedDataLen := 1; len(actual.Data) != expectedDataLen {
+		t.Error(diff.ObjectDiff(expectedDataLen, len(actual.Data)))
```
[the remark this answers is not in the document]
sallyom: doh, good point, thanks :) , done

test/e2e/e2e_test.go
sallyom: dunno, we'll clear that up b4 merging though

sallyom:
@stlaz thanks for the review ;) I've updated and also added a test for updating serving-cert-secret in unit-tests and in extended.  Secret is regenerating for me now, in testing, when I give the existing secret bad data, so that's good.
@mrogers950 not sure if you're still seeing the issue with regeneration you were seeing last week? check to see if this fixes.

pkg/controller/servingcert/controller/secret_updating_controller_test.go
sallyom: I updated the test, added e2e

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -313,6 +313,14 @@ func toRequiredSecret(dnsSuffix string, ca *crypto.CA, service *v1.Service, secr
 		secretCopy.Data = map[string][]byte{}
 	}
 
+	for k, _ := range secretCopy.Data {
```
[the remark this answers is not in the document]
sallyom: k, done

pkg/controller/servingcert/controller/secret_updating_controller_test.go
```diff
@@ -60,6 +60,10 @@ func TestRequiresRegenerationServiceUIDMismatch(t *testing.T) {
 					},
 					OwnerReferences: []metav1.OwnerReference{ownerRef(&v1.Service{ObjectMeta: metav1.ObjectMeta{Name: "foo", UID: types.UID("uid-2")}})},
 				},
+				Data: map[string][]byte{
```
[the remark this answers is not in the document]
sallyom: you are correct

pkg/controller/servingcert/controller/secret_updating_controller_test.go
```diff
@@ -79,6 +83,10 @@ func TestRequiresRegenerationServiceUIDMismatch(t *testing.T) {
 					},
 					OwnerReferences: []metav1.OwnerReference{ownerRef(&v1.Service{ObjectMeta: metav1.ObjectMeta{Name: "foo", UID: types.UID("uid-1")}})},
 				},
+				Data: map[string][]byte{
```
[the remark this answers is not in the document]
sallyom: yup, thanks

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -114,6 +123,20 @@ func (sc *serviceServingCertUpdateController) requiresRegeneration(secret *v1.Se
 		return true, sharedService
 	}
```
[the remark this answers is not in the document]
sallyom: : ) looking at it now, will update

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -309,9 +309,11 @@ func toRequiredSecret(dnsSuffix string, ca *crypto.CA, service *v1.Service, secr
 	if secretCopy.Annotations == nil {
 		secretCopy.Annotations = map[string]string{}
 	}
-	if secretCopy.Data == nil {
-		secretCopy.Data = map[string][]byte{}
-	}
+	// let garbage collector cleanup map allocation, for simplicity
+	secretCopy.Data = map[string][]byte{
+    v1.TLSCertKey: certBytes,
```
mrogers950: A bad indent here

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -114,6 +123,20 @@ func (sc *serviceServingCertUpdateController) requiresRegeneration(secret *v1.Se
...
+	// this checks that the right keys are there
+	// the updateSecret checks the values against known good values
+	for k, _ := range secret.Data {
+		if k != v1.TLSCertKey {
+			if k != v1.TLSPrivateKeyKey {
+				return true, sharedService
+			}
+		}
+	}
+	if len(secret.Data) != 2 {
```
mrogers950: With the change to updateSecret() I don't think we can hit these conditions. updateSecret() makes it so that on update if there is an additional data item or someone fudges with the crt/key, the existing data is kept, so we make them immutable on those conditions.  What you added here wants to regenerate on those conditions. Thinking about regeneration vs. protection of the data, it would be better to have protection for the secret update controller and leave the intentional regeneration to the secret creating controller when you take the action of deleting the secret.

test/e2e/e2e_test.go
```diff
@@ -87,6 +102,88 @@ func pollForServiceServingSecret(client *kubernetes.Clientset, secretName, names
...
+			return false, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	})
+}
+
+func editServiceServingSecretData(client *kubernetes.Clientset, secretName, namespace string) error {
```
mrogers950: This should be addToServiceServingSecretData() and we can make editServiceServingSecretData() a function for another test where we fiddle with the existing data and ensure that the old secret persists.

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: If I edit an existing secret with bad data(anything different), updateSecret() does not result in that change being set back to the old(good) data.  (I'm probably missing something, but the change in updateSecret() is not enough - regenerating is the only way I've found to get back the secret w/ good data).

pkg/controller/servingcert/controller/secret_creating_controller.go
sallyom: fixed : )

pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
```diff
@@ -46,20 +46,21 @@ func (ic *configMapCABundleInjectionController) Sync(obj v1.Object) error {
...
+	// skip updating when the CA bundle is already there
+	if data, ok := configMapCopy.Data[api.InjectionDataKey]; ok &&
+		data == ic.ca && len(configMapCopy.Data) == 1 {
+		return nil
 	}
-
-	configMapCopy.Data[api.InjectionDataKey] = ic.ca
-
-	_, err := ic.configMapClient.ConfigMaps(configMapCopy.Namespace).Update(configMapCopy)
+	configMapCopy.Data = map[string]string{api.InjectionDataKey: ic.ca}
```
stlaz: Do the deep copy before this line instead, you're not modifying anything in the code above.

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -299,18 +302,23 @@ func toRequiredSecret(dnsSuffix string, ca *crypto.CA, service *v1.Service, secr
...
 		return err
 	}
-
 	if secretCopy.Annotations == nil {
 		secretCopy.Annotations = map[string]string{}
 	}
-	if secretCopy.Data == nil {
-		secretCopy.Data = map[string][]byte{}
+	// let garbage collector cleanup map allocation, for simplicity
+	secretCopy.Data = map[string][]byte{
```
stlaz: lines 326, 327 should be removed now

pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
sallyom: done

pkg/controller/servingcert/controller/secret_creating_controller.go
sallyom: yup, : ) fixed

test/e2e/e2e_test.go
sallyom: changed it

pkg/controller/servingcert/controller/secret_updating_controller.go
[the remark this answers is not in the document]
sallyom: updated

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +70,105 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+			return true
 		}
 	}
 	expiry, err := time.Parse(time.RFC3339, expiryString)
 	if err != nil {
-		return true, sharedService
+		return true
 	}
 
 	if time.Now().Add(sc.minTimeLeftForCert).After(expiry) {
```
mrogers950: Use the sc.minTimeLeftForCert or the minTimeLeft argument (using the struct member instead of an extra argument is fine)

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +70,105 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+		}
+		update = true
+	} else {
+		data := map[string][]byte{
+			v1.TLSCertKey:       certBytes,
+			v1.TLSPrivateKeyKey: keyBytes,
+		}
+		// check that we only have the data we care about
+		// if new data does not match, regenerate it
+		if !bytes.Equal(secretCopy.Data[v1.TLSCertKey], data[v1.TLSCertKey]) ||
```
mrogers950: I might be missing something but it seems since you're creating data new each time, it will contain a new cert+key and never match the existing secret, so this path will always lead to regeneration. 

In thinking through the different ways to modify the secret data and what our reaction to it in this else clause should be:
A. Someone adds in an extra data item. We want to get rid of this but keep the same tls.crt and tls.key, so for this we can have a standalone check for len(secretCopy.Data) > 2. If there is more than 2 items, then we make a new data map with the existing tls.crt and tls.key and update.
B. Someone edits tls.crt or tls.key data, replacing them with ascii art. Here we could x509 decode secretCopy data to verify that. If it's been munged then we update with the new cert data. If it's OK then we assume it's the serving cert. If someone replaced the cert data with a legitimate cert unless we had some kind of cache of the current certs issued by the serving CA (or we did an actual verify against the CA) we could not confirm that it belongs.

sallyom: makes sense, I'll update

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: refactored this, checking/replacing/regenerating as needed - my this PR is getting messy

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: updated

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+			}
+			update = true
+		}
+	} else {
+		// if TLSCertKey or TLSPrivateKeyKey missing, replace with valid data
+		// Regenerate the secret
+		if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
+			return false, err
+		}
+		update = true
```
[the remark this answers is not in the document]
sallyom: right, done.

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+}
+
+func (sc *serviceServingCertUpdateController) regenerateSecondCheck(service *v1.Service, secretCopy *v1.Secret) (bool, error) {
+	update := false
+	// If not regenerating, perform checks here to
+	// 1. ensure only the 2 data keys we want exist, TLSCertKey and TLSPrivateKeyKey
+	// 2. ensure cert data is at least a parseable certificate, if not replace with valid data
+	// This does not ensure that somebody didn't swap out secret data for another valid cert.
+	_, ok := secretCopy.Data[v1.TLSCertKey]
+	_, ok2 := secretCopy.Data[v1.TLSPrivateKeyKey]
```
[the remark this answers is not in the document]
sallyom: thanks, done : )

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+	return false
+}
+
+func (sc *serviceServingCertUpdateController) regenerateSecondCheck(service *v1.Service, secretCopy *v1.Secret) (bool, error) {
+	update := false
+	// If not regenerating, perform checks here to
+	// 1. ensure only the 2 data keys we want exist, TLSCertKey and TLSPrivateKeyKey
+	// 2. ensure cert data is at least a parseable certificate, if not replace with valid data
+	// This does not ensure that somebody didn't swap out secret data for another valid cert.
+	_, ok := secretCopy.Data[v1.TLSCertKey]
```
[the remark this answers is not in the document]
sallyom: oh right, duh thanks! : ) done

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+		return true
 	}
 
-	return false, nil
+	return false
+}
+
+func (sc *serviceServingCertUpdateController) regenerateSecondCheck(service *v1.Service, secretCopy *v1.Secret) (bool, error) {
+	update := false
+	// If not regenerating, perform checks here to
```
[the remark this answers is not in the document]
sallyom: ok, done

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
 	if time.Now().Add(sc.minTimeLeftForCert).After(expiry) {
-		return true, sharedService
+		return true
 	}
 
-	return false, nil
+	return false
+}
+
+func (sc *serviceServingCertUpdateController) regenerateSecondCheck(service *v1.Service, secretCopy *v1.Secret) (bool, error) {
```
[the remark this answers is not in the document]
sallyom: yes, you're right, working on that now...

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
-	}
-	if secret.Annotations[api.ServiceUIDAnnotation] != string(sharedService.UID) && secret.Annotations[api.AlphaServiceUIDAnnotation] != string(sharedService.UID) {
-		return false, nil
+		utilruntime.HandleError(fmt.Errorf("unable to get service %s/%s: %v", sharedSecret.Namespace, serviceName, err))
+		return nil
 	}
+	return service
+}
 
+func (sc *serviceServingCertUpdateController) requiresRegenerationFirstCheck(service *v1.Service, secret *v1.Secret, minTimeLeft time.Duration) bool {
```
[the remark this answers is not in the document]
sallyom: done : )

pkg/controller/servingcert/controller/secret_updating_controller.go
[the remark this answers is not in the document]
sallyom: it's good to check, though?, in case someone replaced the cert data with nonsense
(this won't prevent the case where someone for some reason replaces cert data with another valid-ish x509.Cert (ie, it's at least a parseable cert) but will prevent someone from replacing cert data with acsii art)

sallyom:
@mrogers, @Standa, I've added unit-tests for cert replacing/regeneration/removing extra data fields:
For cases where expectedRegeneration = False, I've added: 
* all good, parseable cert, tlsCert/Key data fields - expectReplace = False, expectUpdate = False
* not parseable x509 cert - expectReplace = True, expectUpdate = True
* good cert data, but extra data field - expectReplace = False, expectUpdate = True
* bad data field, no tlsCert/tlsKey field: expectReplace = True, expectUpdate = True

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+		var err error
+		// If not regenerating, perform checks here to
+		// 1. ensure only the 2 data keys we want exist, TLSCertKey and TLSPrivateKeyKey
+		// 2. ensure cert data is at least a parseable certificate, if not replace with valid data
+		// This does not ensure that somebody didn't swap out secret data for another valid cert.
+		update, replace = sc.ensureSecretData(service, secretCopy)
+		if err != nil {
+			return err
+		}
+		if replace {
```
stlaz: This should have stayed inside `ensureSecretData()`. Its presence there should also explain your question in https://github.com/openshift/service-ca-operator/pull/26/files#r270982624 (`toRequiredSecret()` generates valid data).

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,141 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+		update = true
+	}
+	// This ensures someone did not replace secret cert data with ascii art
+	// Check for valid cert data, replace with valid data if not
+	block, _ := pem.Decode([]byte(tlsCert))
+	if block == nil {
+		// Regenerate the secret
+		glog.Infof("Error decoding cert bytes %s from secret: %s namespace: %s, replacing cert", v1.TLSCertKey, secretCopy.Name, secretCopy.Namespace)
+		return true, true
+	} else {
```
stlaz: No need for an `else` as you're returning above, just do it outside any condition scope.

pkg/controller/servingcert/controller/secret_updating_controller_test.go
```diff
@@ -158,9 +178,38 @@ func TestRequiresRegenerationServiceUIDMismatch(t *testing.T) {
...
+					v1.TLSCertKey:       []byte("content"),
+					v1.TLSPrivateKeyKey: []byte("morecontent"),
+				},
 			},
 			expected: false,
 		},
+		{
+			name: "bad data",
```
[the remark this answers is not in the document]
sallyom: done

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: ugh again :(  'elses' fixed.

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: right, but if we didn't call toRequiredSecret, we should check the cert is a cert. Is toRequiredSecret ok separate from ensureSecretData? It makes it convenient to unit-test, to see whether we are correctly replacing the data or not.

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,134 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
 
-	if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
-		return err
+	update := false
+	if sc.requiresRegeneration(service, sharedSecret, sc.minTimeLeftForCert) {
+		// Regenerate the secret
+		if err := toRequiredSecret(sc.dnsSuffix, sc.ca, service, secretCopy); err != nil {
+			return err
+		}
+		update = true
```
sallyom: @stlaz you meant I should update and return here? I do see that.. I'm sorry, this PR is getting difficult to manage.. must merge this today : (

pkg/controller/servingcert/controller/secret_updating_controller.go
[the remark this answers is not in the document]
sallyom: yes, right  I should have, I separated toRequiredSecret from ensureSecretData now, look ok?

pkg/controller/servingcert/controller/secret_updating_controller.go
sallyom: nm I see now, it was above.. I commented there.

pkg/controller/servingcert/controller/secret_updating_controller.go
```diff
@@ -69,70 +72,142 @@ func (sc *serviceServingCertUpdateController) Key(namespace, name string) (metav
...
+		}
+		_, err := sc.secretClient.Secrets(secretCopy.Namespace).Update(secretCopy)
 		return err
 	}
+	// If not regenerating, perform checks here to
+	// 1. ensure only the 2 data keys we want exist, TLSCertKey and TLSPrivateKeyKey
+	// 2. ensure cert data is at least a parseable certificate, if not replace with valid data
+	// This does not ensure that somebody didn't swap out secret data for another valid cert.
+	var err error
+	update, err = sc.ensureSecretData(service, secretCopy)
```
stlaz: Nit: you can scrape `update` and `err` definition and declaration above now and only have `update, err := sc.ensureSecretData(service, secretCopy)`

sallyom: yes, you are correct : ) done

pkg/controller/servingcert/controller/secret_updating_controller_test.go
```diff
@@ -179,21 +302,41 @@ func TestRequiresRegenerationServiceUIDMismatch(t *testing.T) {
...
-				t.Errorf("%s: expected %v, got %v", tc.name, tc.expected, actual)
-			}
-			if service == nil && tc.expected {
+			service := c.getServiceForSecret(tc.secret)
+			if service != nil {
+				isValid := isSecretValidForService(service, tc.secret)
+				if tc.serviceIsValid != isValid {
+					t.Errorf("validService result: %v unexpected", isValid)
+				}
+				if tc.serviceIsValid {
```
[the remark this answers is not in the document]
sallyom: I removed checks for ensureSecretData here altogether.
1) validation prevents one from updating to a secret without the 2 required data keys (tlscert,key)
2) I added e2e tests to check that extra data is removed, secret w/ invalid cert data is regenerated.

pkg/controller/servingcert/controller/secret_updating_controller_test.go
```diff
@@ -179,21 +302,41 @@ func TestRequiresRegenerationServiceUIDMismatch(t *testing.T) {
...
+					minTimeLeft := 1 * time.Hour
+					actualRegen := c.requiresRegeneration(service, tc.secret, minTimeLeft)
+					if tc.expectedRegen != actualRegen {
+						t.Errorf("%s: expectedRegen %v, got %v", tc.name, tc.expectedRegen, actualRegen)
+					}
+					if !tc.expectedRegen {
+						update, actualReplace := c.ensureSecretData(service, tc.secret)
```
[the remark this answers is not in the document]
sallyom: i've updated unit tests, no longer required

PR #28, opened 2019-03-05
e2e: Fix operator pod prefix

mrogers950:
@sallyom

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -7,23 +7,24 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openshift/service-ca-operator/pkg/controller/api"
+	"github.com/openshift/service-ca-operator/pkg/operator/operatorclient"
+
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
-
-	"github.com/openshift/service-ca-operator/pkg/controller/api"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_test.go
```diff
@@ -20,7 +20,7 @@ import (
 const (
 	serviceCAOperatorNamespace   = "openshift-service-ca-operator"
 	serviceCAControllerNamespace = "openshift-service-cert-signer"
-	serviceCAOperatorPodPrefix   = "service-ca-operator"
+	serviceCAOperatorPodPrefix   = "openshift-service-ca-operator"
```
ericavonb: we should import these from the `operatorclient` package: https://github.com/openshift/service-ca-operator/blob/a92546fbaa0c89374c4b4981f57e5b44aae88590/pkg/operator/operatorclient/interfaces.go#L6

mrogers950:
@ericavonb updated

test/e2e/e2e_test.go
```diff
@@ -7,20 +7,21 @@ import (
...
 const (
-	serviceCAOperatorNamespace   = "openshift-service-ca-operator"
-	serviceCAControllerNamespace = "openshift-service-cert-signer"
-	serviceCAOperatorPodPrefix   = "service-ca-operator"
+	serviceCAOperatorNamespace   = operatorclient.OperatorNamespace
+	serviceCAOperatorPodPrefix   = operatorclient.OperatorNamespace // Same as operator namespace
+	serviceCAControllerNamespace = operatorclient.TargetNamespace
 	apiInjectorPodPrefix         = "apiservice-cabundle-injector"
 	configMapInjectorPodPrefix   = "configmap-cabundle-injector"
 	caControllerPodPrefix        = "service-serving-cert-signer"
```
ericavonb: Optional: we can use these too. https://github.com/openshift/service-ca-operator/blob/a92546fbaa0c89374c4b4981f57e5b44aae88590/pkg/controller/api/resourcenames.go#L22-L25

mrogers950: done

PR #29, opened 2019-03-05
replace openshift-service-cert-signer with openshift-service-ca

mrogers950:
@openshift/sig-auth


enj:
Looks like a successful grep.

/lgtm

PR #33, opened 2019-03-14
fix crd yaml

sallyom:
@enj the status was accidentally merged in with the validation.  Similar to here: https://github.com/openshift/cluster-authentication-operator/pull/80#discussion_r262357196


stlaz:
/hold
Fixed in #35 
/close

PR #34, opened 2019-03-15
Add version, status to operator

sallyom:
Add version and status reporting
@enj @mrogers950 

TODO:
bump deps to pull in latest library-go, will do in follow-up PR
w/out this bump, status initiates to Available: True


pkg/operator/operator.go
```diff
@@ -84,16 +91,46 @@ func (c serviceCAOperator) Key() (metav1.Object, error) {
...
+	operatorConfigCopy := operatorConfig.DeepCopy()
 	// This is to push out deployments but does not handle deployment generation like it used to. It may need tweaking.
-	err := sync_v4_00_to_latest(c, operatorConfig)
-	return err
+	syncErr := sync_v4_00_to_latest(c, operatorConfigCopy)
+	if syncErr != nil {
+		c.setFailingStatus(operatorConfigCopy, "OperatorSyncLoopError", syncErr.Error())
+	} else {
+		// Set current version and available status
+		version := os.Getenv("RELEASE_VERSION")
```
derekwaynecarr: can you confirm that you know that all the pods in openshift-service-ca are all running the latest version at this point and no old pods remain?  that is basically what we are trying to ensure when you set the version.

manifests/0000_09_service-ca-operator_07_clusteroperator.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: service-ca
+spec: {}
```
enj: nit: remove

pkg/apis/serviceca/install.go
```diff
@@ -1,26 +0,0 @@
-package kubeapiserver
```
enj: separate delete commit

manifests/0000_09_service-ca-operator_05_deploy.yaml
```diff
@@ -28,6 +28,8 @@ spec:
         env:
         - name: CONTROLLER_IMAGE
           value: quay.io/openshift/origin-service-ca-operator:v4.0
+        - name: RELEASE_VERSION
```
enj: `OPERATOR_IMAGE_VERSION`

pkg/operator/sync_apiservice_v4_00.go
```diff
@@ -63,7 +63,7 @@ func syncAPIServiceController_v4_00_to_latest(c serviceCAOperator, operatorConfi
 	}
 
 	// TODO create a new configmap whenever the data value changes
-	_, configMapModified, err := manageAPIServiceConfigMap_v4_00_to_latest(c.corev1Client, c.eventRecorder, operatorConfig)
+	_, configMapModified, err := manageAPIServiceConfigMap_v4_00_to_latest(c.corev1Client, c.eventRecorder)
```
enj: Keep passing this through and add TODO to combine CLI config and allow overrides with unsupported config overrides.

pkg/operator/sync_signer_v4_00.go
```diff
@@ -66,7 +66,7 @@ func syncSigningController_v4_00_to_latest(c serviceCAOperator, operatorConfig *
 	}
 
 	// TODO create a new configmap whenever the data value changes
-	_, configMapModified, err := manageSigningConfigMap_v4_00_to_latest(c.corev1Client, c.eventRecorder, operatorConfig)
+	_, configMapModified, err := manageSigningConfigMap_v4_00_to_latest(c.corev1Client, c.eventRecorder)
```
enj: Same comment.

pkg/operator/operator.go
sallyom: I've updated the status reporting, will test to be sure it has the correct behavior

manifests/0000_09_service-ca-operator_07_clusteroperator.yaml
sallyom: done

pkg/apis/serviceca/install.go
sallyom: done

manifests/0000_09_service-ca-operator_05_deploy.yaml
sallyom: done

manifests/0000_09_service-ca-operator_07_clusteroperator.yaml
sallyom: removed

pkg/operator/sync_apiservice_v4_00.go
sallyom: follow-up ok?

pkg/operator/sync_signer_v4_00.go
sallyom: follow-up ok?

pkg/operator/operator.go
```diff
@@ -85,15 +93,43 @@ func (c serviceCAOperator) Key() (metav1.Object, error) {
...
+		if setOperatorVersion {
+			version := os.Getenv("OPERATOR_IMAGE_VERSION")
+			if c.versionGetter.GetVersions()["operator"] != version {
+				glog.Infof("Updating clusteroperator %s version: %v", clusterOperatorName, version)
+				// Set current version
+				c.versionGetter.SetVersion("operator", version)
+			}
+		}
+	default:
+		glog.Warningf("unknown custom resource state: %s", operatorConfig.Spec.ManagementState)
```
stlaz: I think we shouldn't even try to check the statuses in this case.

pkg/operator/status.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA, deployments []string) (bool, error) {
+	ready := 0
+	existingDeploymentsAndReplicas := 0
+	var statusErr error
+	for _, dep := range deployments {
+		existing, err := c.appsv1Client.Deployments(operatorclient.TargetNamespace).Get(dep, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			glog.Infof("Deployment %s does not exist", dep)
+			statusMsg := fmt.Sprintf("Deployment %s does not exist", dep)
```
stlaz: I don't think `statusMsg` should appear twice in the glog (once here, once when `Sync()` returns failure). Pick one.

pkg/operator/status.go
```diff
@@ -0,0 +1,151 @@
...
+				ready++
+			}
+		}
+	}
+	// only report version if all deployments are available and updated
+	if ready == len(deployments) {
+		glog.Infof("All deployments managed by service-ca-operator are available and updated")
+		c.setAvailableStatus(operatorConfigCopy)
+		return true, statusErr
+		// set Available if replica instances are created.  Don't report new version if
```
stlaz: This comment should be inside the condition branch it applies to.

pkg/operator/status.go
```diff
@@ -0,0 +1,151 @@
...
+	existingDeploymentsAndReplicas := 0
+	var statusErr error
+	for _, dep := range deployments {
+		existing, err := c.appsv1Client.Deployments(operatorclient.TargetNamespace).Get(dep, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			glog.Infof("Deployment %s does not exist", dep)
+			statusMsg := fmt.Sprintf("Deployment %s does not exist", dep)
+			c.setFailingStatus(operatorConfigCopy, "Error getting service-ca-operator managed deployments", statusMsg)
+			return false, statusErr
+		} else if existing.DeletionTimestamp != nil {
```
stlaz: Each of these if's end with return, no need to do `else if`

pkg/operator/starter.go
```diff
@@ -64,18 +75,28 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+	versionGetter := status.NewVersionGetter()
+	clusterOperator, err := configClient.ConfigV1().ClusterOperators().Get(clusterOperatorName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	for _, version := range clusterOperator.Status.Versions {
+		versionGetter.SetVersion(version.Name, version.Version)
+	}
+	versionGetter.SetVersion("operator", os.Getenv("OPERATOR_IMAGE_VERSION"))
```
stlaz: I know I saw this code in a different repo, too, but won't this just flip the version of `operator` before we perform the checks that we should be updating it?

pkg/operator/operator.go
sallyom: agreed, thanks, done

pkg/operator/status.go
sallyom: removed it here

pkg/operator/status.go
sallyom: fixed

pkg/operator/starter.go
sallyom: yes, it's in cluster-kube-apiserver, pkg/operator/starter.go.  I agree.  I was confused, meant to ask about that, I'll ask them, leave it out for now.  (I added that yesterday late and forgot about it, thanks)

pkg/operator/status.go
sallyom: yes, thanks : ).  I've kept them all as if statements (no else), instead of switch.

PR #35, opened 2019-03-19
Add status back to spec subresource

stlaz:
When service-ca client tries to update our status, it does that via
subresource but the fact that status is the CR's subresource got
unfortunately removed. Add it back.

---------------------------

I believe this should get rid of the log spam.


stlaz:
Adds a bump, all status fields were made optional in the meantime, we were trying to have `generations` empty and that actually caused the logs to grow to 4.5MB

stlaz:
Pushed changes from #34 on top so that we eventually report as `Available`

PR #39, opened 2019-03-28
Fix ERROR: logging before flag.Parse

mrogers950:
@openshift/sig-auth

### DIFF
--- a/cmd/service-ca-operator/main.go
+++ b/cmd/service-ca-operator/main.go
@@ -25,6 +25,8 @@ func main() {
 	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 
+	goflag.CommandLine.Parse([]string{})
+
 	logs.InitLogs()
 	defer logs.FlushLogs()
 


cmd/service-ca-operator/main.go
```diff
@@ -25,6 +25,8 @@ func main() {
 	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 
+	goflag.CommandLine.Parse([]string{})
```
ericavonb: Looking at [k8s.io/apiserver/pkg/util/flag](https://github.com/openshift/service-ca-operator/blob/d966f1916f47ef617a1302c130a1338db8717c31/vendor/k8s.io/apiserver/pkg/util/flag/flags.go#L47), can we replace this and the couple lines above with `utilflag.InitFlags()`?

stlaz: `pflag` is what other library-go based operators are using, let's keep that for standard behavior among them all (fun fact - cluster-openshift-apiserver-operator still uses `NewSSCSCommand()` to init its command.

Using `goflag` all of the sudden surprises me though. Why not use `pflag.CommandLine.Parse()` since you've gone through the trouble to set up that parser on lines 25-26?

stlaz: Ah, parsing the args is already a part of `command.Execute()` :/ However, it's glog which shows the error, so this is but a hack, is it not? Would it break if the user was to set options that affect logging?

mrogers950: This was the workaround that I found via the discussion in https://github.com/kubernetes/kubernetes/issues/17162 that is/was also used in kubernetes . Not sure yet if there is a more permanent fix

PR #40, opened 2019-03-29
bump-deps for co-status-controller fix and update crd validation

sallyom:
* pin/fork kubernetes-apiextensions-apiserver to enable nullable fields/tags in CRDs
* update serviceca crd validation to include nullable fields
* bump deps to kubernetes-1.13


sallyom:
@stlaz AFAICT we need to temporarily pin library-go to 1.12, until we are ready to bump all our deps to kubernetes-1.13.  There is a breaking change in workqueue code `vendor/github.com/openshift/library-go/pkg/controller/metrics/workqueue_metrics.go:58:78: undefined: workqueue.SettableGaugeMetric` that is only fixed by a)pinning library-go to 1.12 or b) updating all of our deps to kubernetes-1.13.  We'll temporarily pin to 1.12 then for now

sallyom:
following CI, looks like we missed cluster-initialization by seconds : ( retesting.

stlaz:
Please remove the line https://github.com/openshift/service-ca-operator/blob/b0e23e6f800bd4cf80620bf83fb817e51310a751/cmd/service-ca-operator/main.go#L28 in your glog to klog commit, it should no longer be necessary (hopefully). It was there as a hack to prevent glog’s “log before flag parse” spamming.

sallyom:
@stlaz removed the goflag line, retesting.

PR #41, opened 2019-04-09
[WIP] Add certificate monitoring

mrogers950:
This uses the monitoring controller annotations (https://github.com/openshift/library-go/pull/323) on secrets and configMaps that the controllers create. Contains a temporary bump until library-go:323 merges.


pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/openshift/service-ca-operator/pkg/boilerplate/controller"
 	"github.com/openshift/service-ca-operator/pkg/controller/api"
+	"github.com/openshift/library-go/pkg/operator/certmonitor"
```
stlaz: The change moved to `github.com/openshift/library-go/pkg/operator/certrotation/metrics`

pkg/controller/configmapcainjector/controller/configmap_injecting_controller.go
```diff
@@ -63,6 +64,7 @@ func (ic *configMapCABundleInjectionController) ensureConfigMapCABundleInjection
 	// make a copy to avoid mutating cache state
 	configMapCopy := current.DeepCopy()
 	configMapCopy.Data = map[string]string{api.InjectionDataKey: ic.ca}
+	certmonitor.SetCertificateMonitoredConfigMap(configMapCopy, certmonitor.CertificateTypeCABundle)
```
stlaz: This is now `LabelAsManagedConfigMap()`

mrogers950:
@stlaz I updated this to use the same cert metrics controller code from library-go (which was recently reverted and not to return until 4.2 (see https://github.com/openshift/library-go/pull/346)). So for now, we use this code in the operator to add metrics only for the service-ca managed certs and configmaps. When the metrics code returns to library-go we can swap out.

PR #42, opened 2019-04-10
[WIP] Auto-renew signing CA

mrogers950:


pkg/operator/sync_common.go
```diff
@@ -81,26 +85,68 @@ func manageControllerResources(c serviceCAOperator, resourcePath string, modifie
 // TODO manage rotation in addition to initial creation
 func manageSignerCA(client coreclientv1.SecretsGetter, eventRecorder events.Recorder) (bool, error) {
 	secret := resourceread.ReadSecretV1OrDie(v4_00_assets.MustAsset("v4.0.0/service-serving-cert-signer-controller/signing-secret.yaml"))
-	_, err := client.Secrets(secret.Namespace).Get(secret.Name, metav1.GetOptions{})
+	var (
+		caCertPem, caKeyPem, signedByOldPem, fullBundlePem []byte
+	)
+
+	existing, err := client.Secrets(secret.Namespace).Get(secret.Name, metav1.GetOptions{})
```
stlaz: How do you like this instead?
```
var certs []*x509.Certificate
if !notFound (
  if err != nil --> return err
  certs, err := parseCurrentCACert
  if err != nil --> return err
)

if len(certs) != 0 && certs[0] != nil {
  cacert, cakey, fullBundlePem, renewed, err := performCrossSign()
  if !renewed || err --> return false, err
  secret.Data["ca-bundle.crt"] = fullBundlePem
  caCertPem = cacert
  caKeyPem, = cakey
} else {
  cacert, cakey, err := createNewSigner()
  if err --> return false, err
  caCertPem = cacert
  caKeyPem, = cakey
}
  secret.Data["tls.crt"] = caCertPem 
  secret.Data["tls.key"] = caKeyPem
  applySecret()
```
IMHO that would be a bit more straightforward implementation, I'd separate the whole condition branches for CrossSign and NewSigner into separate functions as well. I hope I did not miss any detail around

pkg/operator/sync_common.go
```diff
@@ -81,26 +85,68 @@ func manageControllerResources(c serviceCAOperator, resourcePath string, modifie
...
+		if err != nil {
+			return false, err
+		}
+		// If we're not yet half expired, return.
+		certs, err := cert.ParseCertsPEM(existing.Data["tls.crt"])
+		if err != nil {
+			return false, err
+		}
+		if certs[0] != nil {
+			if !util.CertHalfwayExpired(certs[0]) {
```
JAORMX: This seems a little long, isn't it? What about one month? something similar to how certmonger checks before doing a certificate renewal

stlaz: We need a longer period of time so that we can be sure that all payloads using the CA were restarted and thus picked the new bundle

enj:
/uncc

@stlaz @sttts @mfojtik

marun:
Superseded by #73 - please close.

PR #44, opened 2019-04-17
Bug 1700037: Check cert issuer directly

mrogers950:
Since we don't own the annotated service, the signed-by annotation might be overwritten by the owner, leading to continuous cert regeneration. Perform the issuer check by parsing the cert instead of relying on the signed-by annotation.

https://bugzilla.redhat.com/show_bug.cgi?id=1700037

### DIFF
--- a/pkg/controller/servingcert/controller/secret_creating_controller.go
+++ b/pkg/controller/servingcert/controller/secret_creating_controller.go
@@ -13,6 +13,7 @@ import (
 	informers "k8s.io/client-go/informers/core/v1"
 	kcoreclient "k8s.io/client-go/kubernetes/typed/core/v1"
 	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
 	"k8s.io/klog"
 
 	ocontroller "github.com/openshift/library-go/pkg/controller"
@@ -186,7 +187,7 @@ func (sc *serviceServingCertController) requiresCertGeneration(service *corev1.S
 		}
 	}
 
-	_, err := sc.secretLister.Secrets(service.Namespace).Get(secretName)
+	secret, err := sc.secretLister.Secrets(service.Namespace).Get(secretName)
 	if kapierrors.IsNotFound(err) {
 		// we have not created the secret yet
 		return true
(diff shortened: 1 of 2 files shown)


mrogers950:
Updated tests.

stlaz:
This problem makes me wonder - we are currently still relying on the error annotations in service objects to skip error handling when we've done that enough time already. However, in this case, they could also be removed by a different operator owning the service, which may still possibly cause update looping. Should we handle such cases as well? Maybe not set the annotation at all and just handle the error all of the time?

I'm fine with merging this after the comments are fixed, but the above is something we should probably still bear in mind.

ericavonb:
> This problem makes me wonder - we are currently still relying on the error annotations in service objects to skip error handling when we've done that enough time already. However, in this case, the annotations could also be removed by a different operator owning the service, which may still possibly cause update looping. Should we handle such cases as well? Maybe not set the annotation at all and just handle the error all of the time?

I'd have to look into how the controllers are designed here and in library-go, but usually you use a retry counter on the work queue to deal with retry limits.

This makes me wonder if it would be better to use a custom resource (or the CSR resource?) where we can safely record errors in the status, rather than annotations.

mrogers950:
>This makes me wonder if it would be better to use a custom resource (or the CSR resource?) where we can safely record errors in the status, rather than annotations.

This is true about the error annotations that they could also be stomped on. It would probably be better to move this error accounting to a section in the operator status.

sallyom:
> I'm fine with merging this after the comments are fixed,

+1 and follow-up PR/discussion for the error accounting.  this will unblock monitoring team.

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -210,6 +211,24 @@ func (sc *serviceServingCertController) requiresCertGeneration(service *corev1.S
 	return true
 }
 
+// Returns true if the secret certificate has the same issuer common name as the current CA, false
+// if not or if there is a parsing error.
+func (sc *serviceServingCertController) issuedByCurrentCA(secret *corev1.Secret) bool {
```
ericavonb: nit: can be a pure function (remove struct receiver)

ericavonb:
/lgtm
minor comment, can be in follow up though

PR #45, opened 2019-04-18
Bug 1664421: remove emoji from event logs

ericavonb:
Cause: operator used an emoji as the operator informer key

Consequence: error logs contain emoji character, confusing users

Fix: change key to more boring but informative value

Result: errors from operator controller shows "service-ca-operator"
instead of emoji

### DIFF
--- a/pkg/boilerplate/operator/option.go
+++ b/pkg/boilerplate/operator/option.go
@@ -6,10 +6,6 @@ import (
 	"github.com/openshift/service-ca-operator/pkg/boilerplate/controller"
 )
 
-// key is the singleton key shared by all events
-// the value is irrelevant, but pandas are awesome
-const key = "🐼"
-
 type Option func(*operator)
 
 func WithInformer(getter controller.InformerGetter, filter controller.Filter) Option {
@@ -17,7 +13,7 @@ func WithInformer(getter controller.InformerGetter, filter controller.Filter) Op
 		o.opts = append(o.opts,
 			controller.WithInformer(getter, controller.FilterFuncs{
 				ParentFunc: func(obj v1.Object) (namespace, name string) {
-					return key, key // return our singleton key for all events
+					return o.name, o.name // return our singleton key for all events
 				},
 				AddFunc:    filter.Add,
 				UpdateFunc: filter.Update,


pkg/boilerplate/operator/option.go
```diff
@@ -7,8 +7,7 @@ import (
 )
 
 // key is the singleton key shared by all events
-// the value is irrelevant, but pandas are awesome
-const key = "🐼"
+const key = "service-ca-operator"
```
enj: /hold

This is shared code between 3 operators so the value needs to be generic.  I suggest a key value of `熊貓` since the requirement is to remove emoji...

ericavonb: Why does it need to be shared across 3 operators? This value shows up in logs, it needs to be informative, not confuse people.

enj: It literally is shared - the same exact code is copied between the operators - it is library code.

enj: I never got around to trying to move it to lib-go or some other shared spot.

benjaminapetersen: Configurable!  I use it for `console-operator`, but every time I see 🐼 I want 🐹 .  Though really it should be 💩 cuz, you know, if this shows up its broken.

ericavonb: It'd be great to use a library directly; however no matter where the code lives this key has resulted in confusion and bugs being filed. While the code lives here, we need to update it here to something useful.

enj: Here is easy solution: use `o.name`

pkg/boilerplate/operator/option.go
```diff
@@ -6,18 +6,14 @@ import (
...
-
 type Option func(*operator)
 
 func WithInformer(getter controller.InformerGetter, filter controller.Filter) Option {
 	return func(o *operator) {
 		o.opts = append(o.opts,
 			controller.WithInformer(getter, controller.FilterFuncs{
 				ParentFunc: func(obj v1.Object) (namespace, name string) {
-					return key, key // return our singleton key for all events
+					return o.name, o.name // return our singleton key for all events
```
ericavonb: @enj updated to use name

enj:
:coffin: :panda_face: :sob: 

/hold cancel
/lgtm

sferich888:
Can I protest this PR? Post merge.

PR #47, opened 2019-04-22
Use CVO for operator config

damemi:
Use the CVO to manage the operator config, instead of deprecated config helper


ericavonb:
I think we need to run `hack/ update-generated-bindata.sh` to remove it fully. Wish I'd noticed before the tests started passing!

damemi:
Thanks for the catch @ericavonb! Updated and pushed

PR #49, opened 2019-04-25
Shorten signing key secret name

mrogers950:
@openshift/sig-auth 
Shortening this name will be useful because we have one step in the documentation that will reference the secret (force rotate service CA), making it effectively an API.


enj:
In case anyone is wondering why we care about this name, it will be used in the manual rotation docs meaning it will effectively be an API.  Thus it is best to have a name that we can actually say.

PR #51, opened 2019-04-26
WIP: Add webhookconfiguration support

spangenberg:
/hold
/cc @enxebre @bison


pkg/controller/webhookconfigurationcabundle/controller/webhookconfiguration_injecting_controller.go
```diff
@@ -0,0 +1,74 @@
...
+	if !api.HasInjectCABundleAnnotation(sharedValidatingWebhookConfiguration) {
+		return nil
+	}
+
+	// ensure data of validatingwebhookconfiguration
+	return ic.ensureValidatingWebhookConfigurationCABundleInjection(sharedValidatingWebhookConfiguration)
+}
+
+func (ic *webhookConfigurationCABundleInjetorController) ensureValidatingWebhookConfigurationCABundleInjection(current *admissionregistrationv1beta1.ValidatingWebhookConfiguration) error {
+	// skip updating when the CA bundle is already there
```
mrogers950: The client CA bundle can be rotated, so you will need to be sure to allow for it to update.

spangenberg: Thanks for the hint, I'll consider this.

bison:
Just adding a bit more detail. This came out of: https://github.com/openshift/cluster-autoscaler-operator/pull/94

We'd like to avoid granting our operators permission to manage their own webhook configuration objects and instead rely on the CVO and other more privileged operators to do that. The only bit that's missing is having the CA bundle injected into the configuration. For the time being, we're having it injected into a ConfigMap, then copying that into the webhook config ourselves.

pkg/controller/webhookconfigurationcabundle/starter/starter.go
```diff
@@ -0,0 +1,65 @@
...
+
+	scsv1alpha1 "github.com/openshift/api/servicecertsigner/v1alpha1"
+	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/service-ca-operator/pkg/controller/webhookconfigurationcabundle/controller"
+)
+
+func StartWebhookConfigurationCABundleInjector(ctx *controllercmd.ControllerContext) error {
+	// TODO(spangenberg): Adapt to webhookconfiguration - START
+	config := &scsv1alpha1.ConfigMapCABundleInjectorConfig{}
+	if ctx.ComponentConfig != nil {
```
enxebre: why do we need this? where is configCopy used?

enxebre:
overall lgtm, @spangenberg can we get this from WIP to mergeable state?

pkg/controller/webhookconfigurationcabundle/starter/starter.go
spangenberg: Right now I'm just leveraging `ConfigMAPCABundleInjector` config because I was not sure yet what we needed.
This controller is also a general one for Webhooks, there are multiple types... validating, mutating.
I try to put both in here because it makes not too much sense to have multiple.
The code you're questioning will change in the next cycle and then we see wether we need the `configCopy` or not.

spangenberg:
@enxebre the operator controller needs some additions to create a deployment for the `webhookconfiguration-cabundle-injector` which is not in yet. Also I need to create the config object which I mentioned in the previous comment of mine.

pkg/controller/webhookconfigurationcabundle/controller/webhookconfiguration_injecting_controller.go
spangenberg: I removed the check for the moment as you suggested.

pkg/controller/webhookconfigurationcabundle/controller/webhookconfiguration_injecting_controller.go
```diff
@@ -0,0 +1,92 @@
...
+		UpdateFunc: api.HasInjectCABundleAnnotationUpdate,
+	}
+
+	return controller.New("WebhookConfigurationCABundleInjectionController", ic,
+		controller.WithInformer(mutatingWebhookConfigurationInformer, filter),
+		controller.WithInformer(validatingWebhookConfigurationInformer, filter),
+	)
+}
+
+func (ic *webhookConfigurationCABundleInjetionController) Key(_, name string) (metav1.Object, error) {
```
enj: You need distinct sync loops otherwise you will get name collisions.

marun:
I think the approach this PR proposes is fine from a functional perspective, but I'm a bit concerned at supporting more targets for bundle injection via copy-paste development. Future maintenance is likely to be complicated by having multiple identical implementations of the same behavior, and I wonder if the need to support 2 more targets for bundle injection - in addition to the 2 already supported - might be a good opportunity to pay down some tech debt:

 - deprecate `APIServiceCABundleInjectorConfig` and `ConfigMapCABundleInjectorConfig` (thankfully both are still alpha) and replace them with a target-agnostic configuration type (e.g. `CABundleInjectorConfig`).
 - implement a new controller that can target all supported types
   - if necessary, the new configuration type could allow injection to be enabled/disabled on a per-type basis

This strategy should result in a net decrease in the amount of code (common controller) and configuration (only a single set of bindata yaml) that would need to be updated in response to changing requirements, and supporting injection for new types would become trivial.

cc: @sttts

marun:
This PR has been superseded by #78 and #79

PR #53, opened 2019-04-30
README: replace alpha annotations to beta

stlaz:


README.md
```diff
@@ -52,7 +52,7 @@ data:
 kind: ConfigMap
 metadata:
   annotations:
-    service.alpha.openshift.io/inject-cabundle: "true"
+    service.beta.openshift.io/inject-cabundle: "true"
```
adambkaplan: Why are we api-freezing with `beta`? Can't we drop this and make the annotation `service.openshift.io/inject-cabundle`?

enj: Discussed in https://bugzilla.redhat.com/show_bug.cgi?id=1704883#c1

PR #55, opened 2019-04-30
Bug 1703229: Increase leader election values

mrogers950:
These values are not as aggressive and are still reasonable for the CA rotation process' deployment time. 
@openshift/sig-auth 

https://bugzilla.redhat.com/show_bug.cgi?id=1703229


enj:
/hold

Matt will do a bump and remove our explicit setting of these values since they are based on https://github.com/openshift/library-go/pull/361

This is all predicated on the audit logs telling us what is actually causing the writes.

mrogers950:
@enj updated with bump

eparis:
the "authentication" clusteroperator failed with:
`"message": "Progressing: got '404 Not Found' status while trying to GET the OAuth well-known https://10.0.132.43:6443/.well-known/oauth-authorization-server endpoint data",`

eparis:
what that caused by this PR?

mrogers950:
> what that caused by this PR?

I don't think so, looking at the logs there were no cert changes at that time

/retest

PR #57, opened 2019-05-16
bug 1711075: set resource requests on containers

ericavonb:
Prevent pods from eviction, OOMKilling, and CPU starvation when
running in the BestEffort QoS by setting resource requests.

https://github.com/openshift/origin/pull/22787

Limits used:

Memory:
service-ca-operator 80Mi
apiservice-cabundle-injector 50Mi
configmap-cabundle-injector 50Mi
service-serving-cert-signer 120Mi

CPU:
all 10m

bz: https://bugzilla.redhat.com/show_bug.cgi?id=1711075


enj:
Not sure where the numbers come from but they look good enough.

/lgtm

PR #66, opened 2019-08-12
Bug 1740121: Make service-ca have guaranteed scheduling

enj:
Signed-off-by: Monis Khan <mkhan@redhat.com>

### DIFF
--- a/hack/verify-generated-bindata.sh
+++ b/hack/verify-generated-bindata.sh
@@ -4,7 +4,7 @@ source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 function cleanup() {
     return_code=$?
     rm -rf "${TMP_GENERATED_BOOTSTRAP_DIR}"
-    os::test::junit::generate_report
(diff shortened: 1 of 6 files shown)


hack/verify-generated-bindata.sh
```diff
@@ -4,7 +4,7 @@ source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 function cleanup() {
     return_code=$?
     rm -rf "${TMP_GENERATED_BOOTSTRAP_DIR}"
-    os::test::junit::generate_report
+    # os::test::junit::generate_report
```
sallyom: uncomment, right?

enj: `os::test::junit::generate_report` does not exist.  We need to delete all of this and move to the lib-go `make` stuff anyway.

sallyom: oh i see, nm, we need to update CI/update to library-go shell scripts, etc.

PR #68, opened 2019-08-13
Use library-go build scripts

sallyom:
@enj 
This PR adopts library-go's build scripts, gets rid of cruft (hack, tools, .spec, etc).  Makefile follows that of cluster-kube-apiserver-operator.


enj:
Delete `hack`, `tools`, `spec` file, etc.

Makefile
```diff
@@ -1,154 +1,86 @@
...
-# Run e2e tests.
-# 
-# Args:
-#
-# Example:
-#   make test-e2e
-test-e2e:
-	test/e2e.sh
-.PHONY: test-e2e
+all: build
```
enj: This needs to be based off of https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/Makefile

It looks like you used `oc` as your baseline, which has completely distinct requirements since it is a CLI tool.

enj:
Also once this is merged, if you are bored, authn operator needs the same fix :)

Makefile
sallyom: ok, will update, thanks

sallyom:
> Also once this is merged, if you are bored, authn operator needs the same fix :)

yup was planning on it

Makefile
sallyom: updated

sallyom:
once this merges, we'll merge https://github.com/openshift/release/pull/4715 to get `make-verify` running

PR #73, opened 2019-10-08
Automate rotation of the service CA

marun:
This PR is a fork of #42.

Co-authored-by: Matt Rogers <mrogers@redhat.com>
Co-authored-by: Stanislav Laznicka <slaznick@redhat.com>


pkg/operator/util/ca.go
```diff
@@ -0,0 +1,71 @@
...
+// CertHalfwayExpired indicates whether half of the cert validity period has elapsed.
+func CertHalfwayExpired(cert *x509.Certificate) bool {
+	halfValidPeriod := cert.NotAfter.Sub(cert.NotBefore).Nanoseconds() / 2
+	halfExpiration := cert.NotBefore.Add(time.Duration(halfValidPeriod) * time.Nanosecond)
+	return time.Now().After(halfExpiration)
+}
+
+// RotateSigningCA creates a new CA with a set of cross-signed interim CAs to allow for graceful rollover.
+// Returns new CA pem, new CA key pem and the full client CA bundle.
+func RotateSigningCA(currentCert *x509.Certificate, currentKey *rsa.PrivateKey) ([]byte, []byte, []byte, error) {
```
marun: @stlaz Does this look ok?

pkg/operator/util/ca.go
```diff
@@ -0,0 +1,71 @@
...
+	// self-signed as it's signed by the current CA key. This creates a trust bridge between the unrefreshed clients and
+	// refreshed servers as long as refreshed servers serve with a bundle containing this CA and the serving cert.
+	rawSignedByCurrentCert, err := x509.CreateCertificate(crand.Reader, newCert, newCert, newCert.PublicKey, currentKey)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	signedByCurrentCert := &x509.Certificate{Raw: rawSignedByCurrentCert}
+
+	// Create the full CA bundle comprising all 4 certs
+	// TODO(marun) Is the ordering of certs in the bundle important?
```
stlaz: It shouldn't.

pkg/operator/util/ca.go
```diff
@@ -0,0 +1,71 @@
...
+	bundleCerts := []*x509.Certificate{
+		currentCert,
+		signedByNewCert,
```
stlaz: This cert may not even be needed since we have the old CA in the trust bundle. It would be nice to have an e2e test which proves this.

pkg/operator/util/ca.go
```diff
@@ -0,0 +1,71 @@
...
+		signedByCurrentCert,
+		newCert,
+	}
+	bundle, err := cryptutil.EncodeCertificates(bundleCerts...)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	// TODO(marun) Is it necessary to append signedByCurrentCert to
+	// newCertConfig.Certs? PR #42 was doing so.
```
stlaz: See the above comment, I drawed this on our whiteboard and if I am correct, we wouldn't need the `signedByNewCert` instead

pkg/operator/util/ca.go
marun: Done

pkg/operator/sync_common.go
```diff
@@ -78,31 +83,97 @@ func manageControllerResources(c serviceCAOperator, resourcePath string, modifie
...
+		secret.Data["tls.crt"] = caCert
+		secret.Data["tls.key"] = caKey
+	}
+
+	_, mod, err := resourceapply.ApplySecret(client, eventRecorder, secret)
+	return mod, err
+}
+
+// crossSign generates a new CA cert and key
+func crossSign(caCert *x509.Certificate, keyData []byte) ([]byte, []byte, []byte, error) {
```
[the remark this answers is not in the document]
marun: I've inlined this function for now. I'm less concerned with getting the code as neat as possible as getting it functional, and the next tricky thing to tackle appears to be plumbing through the use of the intermediate cert (needs to be appended to service certs if present).

pkg/operator/util/ca.go
```diff
@@ -0,0 +1,71 @@
...
+	newKey := newCertConfig.Key
+
+	// The first interim CA comprises the current CA's public key, private key, and subject. It's self-issued but not
+	// self-signed as it's signed by the new CA key. This creates a trust bridge between refreshed clients and
+	// unrefreshed servers.
+	rawSignedByNewCert, err := x509.CreateCertificate(crand.Reader, currentCert, currentCert, currentCert.PublicKey, newKey)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	signedByNewCert := &x509.Certificate{Raw: rawSignedByNewCert}
```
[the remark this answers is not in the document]
marun: Updated to use ParseCertificate as per your request.  Also, this method now delegates conversion to PEM to the `SigningCA` type to simplify testing.

pkg/operator/util/ca.go
marun: I think the [design doc](https://docs.google.com/document/d/1VqrIERs30M9EyPaqcO43wcSXHXvfe_Ao2W0ThDIohlM/edit) was presuming a signing scheme that relied on intermediates, but since that's not the strategy used by this operator (all service certs are signed by the current 'root'), the only intermediate that matters is #4 (including it with service certs supports support clients that haven't yet updated to the newest ca bundle). That implies that neither intermediate cert needs to be part of the bundle since the old and new certs (`#1` and `#2`) are by themselves sufficient to validate `X` and `Y`. `#3` is unnecessary.

pkg/operator/util/ca.go
marun: I've come to the conclusion that this was never necessary (not sure why #42 was doing so) - CA cert `#4` needs to be included with each new service cert rather than being included with the signing CA.

pkg/operator/util/ca_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+// TestRotateSigningCA validates that service certs signed by pre- and
+// post-rotated CAs can be validated by both the new and old bundles.
+func TestRotateSigningCA(t *testing.T) {
```
marun: @stlaz This test has been rewritten to validate the revised rotation strategy.

marun:
TODO
- [ ] Ensure an intermediate cert (`#4`) is appended to generated signing certs if present
- [ ] Rewrite e2e to validate client trust as per the updated unit test.

pkg/operator/util/ca.go
stlaz: Re-reading the doc and thinking about it once again, that does sound like a correct assumption given the #4 is distributed by the server and #3 will never really sign anything.

pkg/operator/sync_common.go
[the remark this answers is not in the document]
stlaz: That's fine

enj:
/uncc

@stlaz @sttts @mfojtik

pkg/operator/util/ca_test.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	serverAddress := ln.Addr().String()
+	defer ln.Close()
+
+	// Start a server configured with the cert
+	go func() {
+		if err := srv.ServeTLS(ln, "", ""); err != nil && err != http.ErrServerClosed {
+			t.Fatalf("ServeTLS failed: %v", err)
+		}
+	}()
```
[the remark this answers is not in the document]
stlaz: Go merely passes the `srv.Addr` to `net.Listen` so the outcome should be the same - https://golang.org/src/net/http/server.go?s=97796:97864#L3098

I agree that this is a good candidate for an e2e test, but the unit test should eventually have just the verification part (which is what happens on that https server anyway, but I get your point) :+1:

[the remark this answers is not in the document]
marun: The test starts a server on a port guaranteed not to be already bound to ensure reliable execution in a test environment, which is achieved by using an address of `127.0.0.1:0`. As you say, it should be possible to set `srv.Addr` to that address and have the server start on a random port. However, `srv.Addr` is not updated once `ListenAndServeTLS` starts the server - the bound port is only set on the listener internal to that call chain. Hence the reason I'm starting the listener in advance - to enable reliable binding to a random port _and_ retrieval of that port to construct the address the client should target.

In my previous comment I said `I think it would be fine to switch to simple verification before merge`, and by that I meant replacing the https-based testing with cert verification. 

If we wanted to retain https-based testing, I'm not sure why it would be desirable to do that in e2e since verifying the cert chain of trust doesn't depend on a cluster. I'm a firm believer in running tests in the lowest-cost environment possible, and unit testing is about as cheap as we get (this test as currently written has sub-second execution time).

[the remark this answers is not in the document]
marun: I've updated the unit test to perform cert validation, and rewritten the e2e test to validate the trust with http server+client. The e2e test is not yet passing, I'm working on debugging the operator locally.

marun:
This PR is dependent on https://github.com/openshift/library-go/pull/560 and https://github.com/openshift/api/pull/482.

The code is functionally complete, but the architecture of this operator make it very challenging to debug locally. For now I'm going to rely on CI.

pkg/operator/sync_common.go
```diff
@@ -78,31 +85,88 @@ func manageControllerResources(c serviceCAOperator, resourcePath string, modifie
...
+		key, err := keyutil.ParsePrivateKeyPEM(keyData)
+		if err != nil {
+			return false, err
+		}
+		signingCA, err := util.RotateSigningCA(existingCert, key.(*rsa.PrivateKey))
+		if err != nil {
+			return false, err
+		}
+		d := secret.Data
+		d[corev1.TLSCertKey], d[corev1.TLSPrivateKeyKey], d[api.IntermediateDataKey], d[api.BundleDataKey], err = signingCA.GetPEMBytes()
```
stlaz: You swapped Intermediate and Bundle keys here

pkg/operator/util/ca.go
```diff
@@ -0,0 +1,78 @@
...
+const SigningCertificateLifetimeInDays = 365 // 1 year
+
+type SigningCA struct {
+	Config             *crypto.TLSCertificateConfig
+	Bundle             []*x509.Certificate
+	IntermediateCACert *x509.Certificate
+}
+
+// GetPEMBytes returns PEM-encodings of the CA cert, key, bundle and intermediate CA cert.
+func (ca *SigningCA) GetPEMBytes() ([]byte, []byte, []byte, []byte, error) {
```
stlaz: Inline this within `RotateSigningCA()`, the struct above is single-purpose, you're only carrying the bits over.

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -313,11 +322,17 @@ func getServingCert(dnsSuffix string, ca *crypto.CA, service *corev1.Service) (*
...
+	if intermedicateCACert != nil {
+		servingCert.Certs = append(servingCert.Certs, intermedicateCACert)
+	}
 	return servingCert, nil
 }
 
-func toRequiredSecret(dnsSuffix string, ca *crypto.CA, service *corev1.Service, secretCopy *corev1.Secret) error {
-	servingCert, err := getServingCert(dnsSuffix, ca, service)
+func toRequiredSecret(dnsSuffix string, ca *crypto.CA, intermediateCACert *x509.Certificate, service *corev1.Service, secretCopy *corev1.Secret) error {
+	servingCert, err := getServingCert(dnsSuffix, ca, intermediateCACert, service)
```
stlaz: The original PR was not adding the intermediate to the serving cert. I think commonly the server just sends its whole chain. I wonder whether this step would actually make a server which would otherwise send just its own cert send both or not.

pkg/controller/servingcert/starter/starter.go
```diff
@@ -31,6 +35,18 @@ func StartServiceServingCertSigner(ctx *controllercmd.ControllerContext) error {
 		return err
 	}
 
+	var intermediateCert *x509.Certificate
+	if len(config.IntermediateCertFile) > 0 {
```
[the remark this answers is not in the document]
stlaz: Nvm, this is wrong

glide.diff
```diff
@@ -0,0 +1,72 @@
+diff --no-dereference -N -r current/vendor/github.com/openshift/api/servicecertsigner/v1alpha1/types.go updated/vendor/github.com/openshift/api/servicecertsigner/v1alpha1/types.go
```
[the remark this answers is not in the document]
marun: This was meant to be pushed - the commits associated with it are marked `DO NOT MERGE` because as per [my previous comment](https://github.com/openshift/service-ca-operator/pull/73#issuecomment-543607442) changes need to merge to deps before this PR will be mergeable.

pkg/operator/sync_common.go
marun: Good catch!  Done.

pkg/operator/util/ca.go
marun: Externalizing conversion to this method simplifies unit testing since conversion to PEM is unnecessary for the purposes of cert verification. Or would you like to see the exported method `RotateSigningCA` return PEM and an internal method `rotateSigningCA` that returns `SigningCA`?

pkg/controller/servingcert/controller/secret_creating_controller.go
marun: The server only sends the certs that are supplied to it. Typically that won't include an entire trust chain, but only the certs required for the client to verify from what the client already knows (i.e. a ca bundle). The goal here is ensuring that a server using a serving cert will send clients `#4` and `Y` so that unrefreshed clients will be able to verify `Y` against an old bundle that contains `#1` but is missing `#2` (because `Y` -> `#4` -> `#1`).

I talked to Matt yesterday and confirmed that the original PR was in error in not adding the intermediate (`#4`) to new serving certs. The design doc even explicitly says it is required - `Distribute #4 with the new service-certs (appended to PEM)`.

I've also confirmed this in the unit testing - try removing [the addition of the intermediate cert](https://github.com/openshift/service-ca-operator/pull/73/files#diff-c482e5ba51661f6d2b95bcc56e0cd468R59) and the `Server rotated` test scenario will fail as expected.

marun:
@stlaz The intermediate cert (`#4`) is only required to be included in serving certs - this ensures that clients without the newest ca bundle can verify trust. Clients with the updated ca bundle containing `#1` and `#2` don't need `#4` since both `X` and `Y` can be validated without it.

pkg/operator/util/ca.go
marun: Going back to this comment, how do you imagine that `#4` is distributed by the server if it is not appended to generated serving certs?

marun:
@stlaz Thank you for your diligent review, much appreciate!

> I'm not sure about the intermediate CA split from the bundle.

The intermediate cert (`#4`) is only required to be included in serving certs - this ensures that clients without the newest ca bundle can verify trust. Clients with the updated ca bundle containing `#1` and `#2` don't need `#4` since both `X` and `Y` can be validated without it.

marun:
Now that [the enhancement](https://github.com/openshift/enhancements/pull/83) as merged, this PR is ready for review. I've merged #77 (forced rotation) into this PR since it ended up cleaning up the code in ways that I thought were more reviewable.

As previously commented, openshift/library-go#560 and openshift/api#482 need to merge before this PR will be mergeable. This PR is vendoring those PR branches instead of master until master includes the necessary changes. 

Once the dep PRs merge, should a bump be performed separately from this PR?

test/e2e/e2e_test.go
```diff
@@ -11,15 +11,23 @@ import (
 	"testing"
 	"time"
 
-	"github.com/openshift/service-ca-operator/pkg/controller/api"
-	"github.com/openshift/service-ca-operator/pkg/operator/operatorclient"
-
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
```
stlaz: nit: `kruntime`?

pkg/operator/sync_common.go
```diff
@@ -78,31 +82,79 @@ func manageControllerResources(c serviceCAOperator, resourcePath string, modifie
...
+			return false, nil
+		}
+		// Ensure the updated existing secret is applied below
+		secret = existing
+	}
+
+	_, mod, err := resourceapply.ApplySecret(client, eventRecorder, secret)
+
+	if err == nil && len(rotationMsg) > 0 {
+		eventRecorder.Eventf("RotatedServiceCA", rotationMsg)
```
stlaz: nit: `ServiceCARotated` matches better other event names

test/util/rotate.go
```diff
@@ -0,0 +1,171 @@
...
+	srv := http.Server{}
+	// Start a server configured with the cert and key
+	go func() {
+		if err := srv.ServeTLS(ln, certFile.Name(), keyFile.Name()); err != nil && err != http.ErrServerClosed {
+			t.Fatalf("ServeTLS failed: %v", err)
+		}
+	}()
+	defer func() {
+		err = srv.Close()
+		if err != nil {
```
stlaz: nit: `if err := srv.Close(); err != nil`, unless you meant to assign to the outer scope error

pkg/operator/rotate_test.go
```diff
@@ -0,0 +1,280 @@
...
+	// Create a brand new signing secret
+	newCAConfig, err := crypto.MakeSelfSignedCAConfig("foo", signingCertificateLifetimeInDays)
+	if err != nil {
+		t.Fatalf("error generating a new ca: %v", err)
+	}
+	newSigningSecret, newCACert := caToSigningSecret(t, newCAConfig)
+
+	// Create a secret that has been force-rotated for a reason
+	forceRotatedSigningSecret := newSigningSecret.DeepCopy()
+	recordForcedRotationReason(forceRotatedSigningSecret, "42")
```
stlaz: lol

pkg/operator/sync_common.go
```diff
@@ -177,7 +177,9 @@ func manageConfigMapCABundleControllerDeployment(client appsclientv1.AppsV1Inter
 func manageDeployment(client appsclientv1.AppsV1Interface, eventRecorder events.Recorder, options *operatorv1.ServiceCA, resourcePath string, forceDeployment bool) (bool, error) {
 	required := resourceread.ReadDeploymentV1OrDie(v4_00_assets.MustAsset(resourcePath + "deployment.yaml"))
 	required.Spec.Template.Spec.Containers[0].Image = os.Getenv("CONTROLLER_IMAGE")
-	required.Spec.Template.Spec.Containers[0].Args = append(required.Spec.Template.Spec.Containers[0].Args, fmt.Sprintf("-v=%s", options.Spec.LogLevel))
+	if len(options.Spec.LogLevel) > 0 {
+		required.Spec.Template.Spec.Containers[0].Args = append(required.Spec.Template.Spec.Containers[0].Args, fmt.Sprintf("-v=%s", options.Spec.LogLevel))
```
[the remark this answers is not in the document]
marun: Done

manifests/03_operator.cr.yaml
```diff
@@ -6,3 +6,5 @@ metadata:
     release.openshift.io/create-only: "true"
 spec:
   managementState: Managed
+  # Configure the log level of managed deployments
+  logLevel: "4"
```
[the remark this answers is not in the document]
marun: > With the change from previous comment, you should not need this, I hope. However, it changes the default loglevel to 2.

Done
 
> Isn't it weird we're running with debug loglevel by default?

I guess so. I wasn't intending to affect the logging behavior in this PR, not sure why previous maintainers decided to set 4 as the default.

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -227,7 +231,21 @@ func (sc *serviceServingCertController) issuedByCurrentCA(secret *corev1.Secret)
...
+	}
+
+	certAuthorityKeyId := certs[0].AuthorityKeyId
+	caSubjectKeyId := sc.ca.Config.Certs[0].SubjectKeyId
+	// Use key identifier chaining if the SubjectKeyId is populated in the CA
+	// certificate. AuthorityKeyId may not be set in the serving certificate if it was
+	// generated before serving cert generation was updated to include the field.
+	checkKeyIdentifierChaining := len(caSubjectKeyId) > 0
+	if checkKeyIdentifierChaining {
+		return bytes.Compare(certAuthorityKeyId, caSubjectKeyId) == 0
```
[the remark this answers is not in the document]
marun: Done

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -301,23 +319,34 @@ func toBaseSecret(service *corev1.Service) *corev1.Secret {
 	}
 }
 
-func getServingCert(dnsSuffix string, ca *crypto.CA, service *corev1.Service) (*crypto.TLSCertificateConfig, error) {
-	dnsName := service.Name + "." + service.Namespace + ".svc"
+func GetServingCert(dnsSuffix string, ca *crypto.CA, intermediateCACert *x509.Certificate, objMeta *metav1.ObjectMeta) (*crypto.TLSCertificateConfig, error) {
```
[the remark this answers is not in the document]
marun: Done

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -301,23 +319,34 @@ func toBaseSecret(service *corev1.Service) *corev1.Secret {
...
 	servingCert, err := ca.MakeServerCert(
 		sets.NewString(dnsName, fqDNSName),
 		certificateLifetime,
-		cryptoextensions.ServiceServerCertificateExtensionV1(service),
+		cryptoextensions.ServiceServerCertificateExtensionV1(objMeta.UID),
 	)
 	if err != nil {
 		return nil, err
 	}
+	if len(servingCert.Certs) > 1 {
```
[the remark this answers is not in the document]
marun: Done.

It's not really a precaution, it's just discontinuing the practice of unnecessarily including the issuing cert. Maybe this is desirable if the issuing cert were an intermediate that was not shipped with the CA bundle, but that isn't the case for service CA. I'm guessing this is a consequence of library-go serving many different use cases, one of which requires this behavior.

pkg/operator/rotate.go
```diff
@@ -0,0 +1,217 @@
...
+
+type signingCA struct {
+	config             *crypto.TLSCertificateConfig
+	bundle             []*x509.Certificate
+	intermediateCACert *x509.Certificate
+	oldCAExpiry        time.Time
+}
+
+// getPEMBytes returns PEM-encodings of the CA cert, key, bundle and intermediate CA cert.
+func (ca *signingCA) getPEMBytes() ([]byte, []byte, []byte, []byte, error) {
```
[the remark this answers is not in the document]
marun: I've refactored this method to `updateSigningSecret` which avoids the need to return so many values.

manifests/03_operator.cr.yaml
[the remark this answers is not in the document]
marun: I've updated the verbosity of logging of rotation initiation to `2`, since it is infrequent and is something that the user should be informed of even without debug logging enabled.

test/e2e/e2e_test.go
marun: Done

`pkgruntime` is the alias I've commonly seen in k/k and related upstream projects, I guess `kruntime` is the openshift version?

pkg/operator/sync_common.go
marun: Done

test/util/rotate.go
marun: Done

pkg/operator/sync_common.go
```diff
@@ -78,31 +82,79 @@ func manageControllerResources(c serviceCAOperator, resourcePath string, modifie
...
+
+	rotationMsg := ""
+	if existingCert == nil {
+		// Secret does not exist or lacks the expected cert.
+		err = initializeSigningSecret(secret)
+		if err != nil {
+			return false, err
+		}
+	} else {
+		rotationMsg, err = rotateSigningSecret(existing, existingCert, rawUnsupportedServiceCAConfig)
```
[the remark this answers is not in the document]
marun: Why is it necessary to write the changes to `secret`? In the case of an existing secret, the expected form was previously only used if the secret did not exist or had a missing/malformed value for `tls.crt`.

`existing` is the result of a direct client call rather than being sourced from an informer, so there is no cache involved.

marun:
@stlaz Thank you for your diligent review of this admittedly huge PR!

I've addressed most of your comments (with the exception of [one](https://github.com/openshift/service-ca-operator/pull/73#discussion_r341094150) that I would appreciate further feedback on) in a fixup commit to simplify reviewing. The exception was for some of the minor supporting commits, which I amended.

Note that the `Fixup 2` commit addresses an oversight that seems obvious in hindsight. Rotating a CA near expiry (e.g. for a 4.1 cluster upgraded to 4.3 very close to a year after deployment) previously might have resulted in the old CA not being trusted long enough for clients and servers to refresh. `Fixup 2` ensures that the old CA remains trusted for a minimum duration.

@sttts ^ I think you touched on this in your review of the enhancement, let me know what you think about the proposed fix.

pkg/operator/sync_common.go
```diff
@@ -177,7 +178,7 @@ func manageConfigMapCABundleControllerDeployment(client appsclientv1.AppsV1Inter
 func manageDeployment(client appsclientv1.AppsV1Interface, eventRecorder events.Recorder, options *operatorv1.ServiceCA, resourcePath string, forceDeployment bool) (bool, error) {
 	required := resourceread.ReadDeploymentV1OrDie(v4_00_assets.MustAsset(resourcePath + "deployment.yaml"))
 	required.Spec.Template.Spec.Containers[0].Image = os.Getenv("CONTROLLER_IMAGE")
-	required.Spec.Template.Spec.Containers[0].Args = append(required.Spec.Template.Spec.Containers[0].Args, fmt.Sprintf("-v=%s", options.Spec.LogLevel))
+	required.Spec.Template.Spec.Containers[0].Args = append(required.Spec.Template.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", loglevel.LogLevelToKlog(options.Spec.LogLevel)))
```
sttts: This is not only about breakage with empty value, but about the API semantics. Before we wanted numbers as string type, now Normal, Debug, Trace, TraceAll.

Was this just broken before?

sttts: Looking at the API, it was broken.

manifests/06_config.yaml
```diff
@@ -1,10 +0,0 @@
-# TODO: axe this file in favor of EnsureOperatorConfigExists within the operator
-apiVersion: operator.openshift.io/v1
-kind: ServiceCA
-metadata:
-  name: cluster
-spec:
-  managementState: Managed
-  version: 4.0.0
-  logLevel: "4"
-  replicas: 1
```
sttts: nit: confusing commit message. This one is a duplicate. The other one existing has `release.openshift.io/create-only: "true"` set.

pkg/controller/servingcert/controller/secret_creating_controller.go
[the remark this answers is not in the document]
sttts: The primary CA is not trusted (i.e. signed by some root CA which everybody trusts), right? So every client should have access to the CA bundle already, and hence this change should not have any consequence?

glide.yaml
```diff
@@ -15,11 +15,15 @@ import:
...
+  # TODO(marun) Update to master once this branch has merged
+  version: add-intermediate-cert-file
+  repo: git@github.com:marun/api.git
 - package: github.com/openshift/client-go
   version: master
 - package: github.com/openshift/library-go
-  version: master
+  # TODO(marun) Update to master once this branch has merged
+  version: add-akid-skid-to-certs
+  repo: git@github.com:marun/library-go.git
```
sttts: tagged. Should merge any moment.

glide.yaml
```diff
@@ -15,11 +15,15 @@ import:
 - package: k8s.io/component-base
   version: kubernetes-1.16.0
 - package: github.com/openshift/api
-  version: master
+  # TODO(marun) Update to master once this branch has merged
+  version: add-intermediate-cert-file
+  repo: git@github.com:marun/api.git
```
sttts: tagged. Should merge any moment.

pkg/controller/api/api.go
```diff
@@ -7,9 +7,12 @@ import (
...
-	InjectCABundleAnnotationName      = "service.beta.openshift.io/inject-cabundle"
-	AlphaInjectCABundleAnnotationName = "service.alpha.openshift.io/inject-cabundle"
-	InjectionDataKey                  = "service-ca.crt"
+	InjectCABundleAnnotationName       = "service.beta.openshift.io/inject-cabundle"
+	AlphaInjectCABundleAnnotationName  = "service.alpha.openshift.io/inject-cabundle"
+	ForcedRotationReasonAnnotationName = "service-ca.operators.openshift.io/forced-rotation-reason"
+	InjectionDataKey                   = "service-ca.crt"
+	BundleDataKey                      = "ca-bundle.crt"
+	IntermediateDataKey                = "intermediate-ca.crt"
 )
```
sttts: oneliner godocs would be helpful

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -227,7 +231,20 @@ func (sc *serviceServingCertController) issuedByCurrentCA(secret *corev1.Secret)
 		return false
 	}
 
-	return certs[0].Issuer.CommonName == sc.commonName()
+	if certs[0].Issuer.CommonName != sc.commonName() {
```
sttts: this is for backwards compatibility? godoc please.

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -301,23 +318,34 @@ func toBaseSecret(service *corev1.Service) *corev1.Secret {
 	}
 }
 
-func getServingCert(dnsSuffix string, ca *crypto.CA, service *corev1.Service) (*crypto.TLSCertificateConfig, error) {
-	dnsName := service.Name + "." + service.Namespace + ".svc"
+func GetServingCert(dnsSuffix string, ca *crypto.CA, intermediateCACert *x509.Certificate, serviceObjectMeta *metav1.ObjectMeta) (*crypto.TLSCertificateConfig, error) {
```
sttts: this is not a getter, but a constructor. NewServingCert

sttts: or MakeServingCert

pkg/controller/servingcert/starter/starter.go
```diff
@@ -31,6 +37,11 @@ func StartServiceServingCertSigner(ctx *controllercmd.ControllerContext) error {
 		return err
 	}
 
+	intermediateCACert, err := readIntermediateCACert(config.IntermediateCertFile)
```
sttts: nit: would prefer to have the `len == 0` logic here. Noop funcs with empty parameters are always suspicious.

pkg/controller/servingcert/starter/starter.go
```diff
@@ -64,3 +77,29 @@ func StartServiceServingCertSigner(ctx *controllercmd.ControllerContext) error {
...
+// that no certificate was specified.
+func readIntermediateCACert(filename string) (*x509.Certificate, error) {
+	if len(filename) == 0 {
+		return nil, errors.New("filename not provided")
+	}
+
+	certsPEMBlock, err := ioutil.ReadFile(filename)
+	if os.IsNotExist(err) {
+		klog.V(4).Infof("%q does not exist which indicates that an intermediate certificate was not specified", filename)
+		return nil, nil
```
sttts: shouldn't we fall over instead? When can this happen?

pkg/controller/servingcert/starter/starter.go
```diff
@@ -64,3 +77,29 @@ func StartServiceServingCertSigner(ctx *controllercmd.ControllerContext) error {
...
+	}
+	if err != nil {
+		return nil, err
+	}
+	certs, err := cert.ParseCertsPEM(certsPEMBlock)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing intermediate cert from %s: %s", filename, err)
+	}
+	if len(certs) > 1 {
```
sttts: `!= 1`

pkg/operator/rotate.go
```diff
@@ -0,0 +1,242 @@
...
+
+// forcedRotationRequired indicates whether the force rotation reason is not empty and
+// does not match the annotation stored on the signing secret.
+func forcedRotationRequired(secret *corev1.Secret, reason string) bool {
+	if len(reason) == 0 {
+		return false
+	}
+	if secret.Annotations == nil {
+		return true
+	}
```
sttts: nit: you can skip this if

pkg/operator/rotate.go
```diff
@@ -0,0 +1,242 @@
...
+	d[corev1.TLSCertKey] = caPEM
+	d[corev1.TLSPrivateKeyKey] = keyPEM
+	d[api.BundleDataKey] = bundlePEM
+	d[api.IntermediateDataKey] = intermediatePEM
+
+	return nil
+}
+
+// rotateSigningSecret rotates the CA provided by the given secret and updates the secret
+// with the new CA's data.
```
sttts: ... or does nothing if neither a force rotation takes place nor the old ca expired.

pkg/operator/rotate.go
```diff
@@ -0,0 +1,242 @@
...
+func rotateSigningSecret(secret *corev1.Secret, currentCACert *x509.Certificate, rawUnsupportedServiceCAConfig []byte) (string, error) {
```
sttts: maybeRotateSigningSecret

pkg/operator/rotate.go
```diff
@@ -0,0 +1,242 @@
...
+		klog.V(2).Infof("Rotating service CA due to the CA being past the mid-point of its validity.")
+	}
+
+	keyData := secret.Data[corev1.TLSPrivateKeyKey]
+	if len(keyData) == 0 {
+		return "", fmt.Errorf("signing secret is missing a value for %q", corev1.TLSPrivateKeyKey)
+	}
+	key, err := keyutil.ParsePrivateKeyPEM(keyData)
+	if err != nil {
+		return "", err
```
sttts: better error context?

pkg/operator/rotate.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	signingCA, err := rotateSigningCA(currentCACert, key.(*rsa.PrivateKey))
```
sttts: could this be a non RSA key?

pkg/operator/rotate.go
```diff
@@ -0,0 +1,242 @@
...
+		return nil, fmt.Errorf("error creating intermediate CA certificate: %v", err)
+	}
+
+	return caCert, nil
+}
+
+// certHalfwayExpired indicates whether half of the cert validity period has elapsed.
+func certHalfwayExpired(cert *x509.Certificate) bool {
+	halfValidPeriod := cert.NotAfter.Sub(cert.NotBefore).Nanoseconds() / 2
+	halfExpiration := cert.NotBefore.Add(time.Duration(halfValidPeriod) * time.Nanosecond)
```
sttts: is that `* time.Nanosecond`  really needed? Doesn't Add take a time.Time?

pkg/operator/rotate.go
```diff
@@ -0,0 +1,242 @@
...
+	return time.Now().After(halfExpiration)
+}
+
+// forceRotationReason attempts to retrieve a force rotation reason
+// from the provided raw UnsupportedConfigOverrides.
+func forceRotationReason(rawUnsupportedServiceCAConfig []byte) (string, error) {
+	serviceCAConfig := &unsupportedServiceCAConfig{}
+	if raw := rawUnsupportedServiceCAConfig; len(raw) > 0 {
+		jsonRaw, err := kyaml.ToJSON(raw)
```
sttts: I believe this is JSON already.

pkg/operator/sync_common.go
```diff
@@ -78,31 +83,79 @@ func manageControllerResources(c serviceCAOperator, resourcePath string, modifie
...
+			if len(certs) > 0 {
+				existingCert = certs[0]
+			}
+		}
 	}
+
+	rotationMsg := ""
+	if existingCert == nil {
+		// Secret does not exist or lacks the expected cert.
+		err = initializeSigningSecret(secret)
```
sttts: `if err := ... ; err !=nil {`

pkg/operator/rotate.go
sttts: At those which come from secrets externally should have a type check.

pkg/operator/sync_common.go
marun: Is this comment actionable?

manifests/06_config.yaml
marun: Done

pkg/operator/sync_common.go
sttts: No. You are fixing the feature.

pkg/controller/api/api.go
marun: Done

pkg/controller/servingcert/controller/secret_creating_controller.go
marun: I'm assuming you meant add a comment, since godoc applies to declarations? Or did you want a change in godoc of the function?

pkg/controller/servingcert/controller/secret_creating_controller.go
marun: Done

pkg/controller/servingcert/starter/starter.go
marun: Done

pkg/controller/servingcert/starter/starter.go
marun: An intermediate cert will only exist if rotation has performed. For a new cluster, that will not be true. I've updated the godoc for the function to reflect this, PTAL.

pkg/controller/servingcert/starter/starter.go
marun: I'm afraid I'm not sure I understand what you mean by `Noop funcs with empty parameters`.  I've added a comment indicating that the intermediate cert will only be expected after rotation, does that help?

pkg/operator/rotate.go
marun: Will `Annotations` will never be nil on a resource returned from the API?  In unit testing that won't be true unless the map is explicitly initialized, but if that's testing-only I should update initialization of secrets used for testing.

pkg/operator/rotate.go
marun: Done

pkg/operator/rotate.go
marun: Done

pkg/operator/rotate.go
marun: Done

pkg/operator/rotate.go
marun: Done

pkg/operator/rotate.go
sttts: `nil["foo"]` gives empty string. Doesn't that match your logic here?

pkg/controller/servingcert/starter/starter.go
sttts: would check for the filename to be non-empty here, not inside of a func (which has no clue about the context it is called in).

pkg/operator/rotate.go
marun: I've removed this function in preference to triggering time-based rotation when the current CA's expiry provides less than the minimum trust duration (~6 months) rather than checking that the CA's is more than half-way through its validity period. This is in keeping with how long the previous CA will be trusted. WDYT?

pkg/operator/rotate.go
marun: Done

pkg/operator/sync_common.go
marun: Done

pkg/controller/servingcert/starter/starter.go
marun: Done

pkg/operator/rotate.go
marun: I was confused - I added this check some time ago in response to an error I saw but it was clearly unrelated.

Done.

marun:
@sttts: Addressed your comments with https://github.com/openshift/service-ca-operator/pull/73/commits/905f5b0251451784567a79b9c4a73b04d53a89c9 and https://github.com/openshift/service-ca-operator/pull/73/commits/fcadd10bc374f5d9e49965347460599327e4c816.

pkg/controller/servingcert/controller/secret_creating_controller.go
sttts: yes

pkg/operator/rotate.go
sttts: it's the same at the end, isn't it?

marun: Effectively the same, but I think it's more clear to using the same constant (`minimumTrustDuration`) for both detection that time-based rotation is required and to ensure that the previous CA remains trusted for at least that same duration.

pkg/controller/servingcert/controller/secret_creating_controller.go
marun: Done

PR #76, opened 2019-10-23
status: do not flap degraded from false to unknown

mfojtik:
In case the Degraded condition is not set to False, the union will set it as Degraded=Unknown with NoData as reason (because we haven't said whether we are or not degraded).

This will always set degraded to false when there is no error returned from the loop.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -100,9 +100,7 @@ func (c serviceCAOperator) Sync(obj metav1.Object) error {
 		if err != nil {
 			setDegradedTrue(operatorConfigCopy, "OperatorSyncLoopError", err.Error())
 		} else {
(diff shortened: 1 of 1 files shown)


mfojtik:
@deads2k @marun https://github.com/openshift/library-go/pull/563 might be related to this, but I think we should always set degraded=false, regardless of whether we were degraded=true before or not if the sync loop ran successfully without errors, WDYT?

marun:
> @deads2k @marun [openshift/library-go#563](https://github.com/openshift/library-go/pull/563) might be related to this, but I think we should always set degraded=false, regardless of whether we were degraded=true before or not if the sync loop ran successfully without errors, WDYT?

I don't actually understand the problem this PR is intended to solve. How could I best educate myself?

stlaz:
/lgtm

I forgot about this PR.

It looks reasonable, the check whether status really needs to be updated happens later in the helper function, so there is no reason to be afraid to just always set it.

PR #77, opened 2019-10-25
WIP Enable forced rotation

marun:
**NOTE** This PR is not ready for review until #73 has merged

This PR adds support for forcing service CA rotation by setting `spec.forcefulRotation.reason` in the operator config. This is in addition to the time-based rotation implemented in #73.

This is the second PR implementing https://github.com/openshift/enhancements/pull/83

cc: @sttts @deads2k @stlaz @mrogers950


manifests/03_operator.cr.yaml
```diff
@@ -1,3 +1,4 @@
+# TODO: axe this file in favor of EnsureOperatorConfigExists within the operator
```
sttts: don't think this TODO is valid

manifests/03_operator.cr.yaml
```diff
@@ -6,3 +7,6 @@ metadata:
     release.openshift.io/create-only: "true"
 spec:
   managementState: Managed
+  version: 4.0.0
+  logLevel: "4"
+  replicas: 1
```
sttts: are these really existing in the API ? look odd.

marun: I'm assuming this is legacy. I'm just copying over from the deleted `06_config.yaml` since its subsequent application was overriding this one. Since these fields do not seem to be present in operator CRs in other repos, I'll assume it can be removed.

manifests/03_operator.cr.yaml
marun: Done

manifests/03_operator.cr.yaml
marun: Removed.

marun:
The code in this PR was sufficiently refactored from #73 that I thought it best to fold the proposed into that PR.

PR #80, opened 2019-11-06
WIP Check rotation breakage

marun:
This PR is not intended to be reviewed or merged - it's a testbed for checking for operators that do not refresh their serving cert or ca bundle after a rotation occurs.


marun:
May revisit in the future, but closing for now.

PR #81, opened 2019-11-07
Enable Auth for the metrics endpoints for the controllers

sohankunkerkar:


bindata/v4.0.0/apiservice-cabundle-controller/defaultconfig.yaml
```diff
@@ -1,7 +1,3 @@
 apiVersion: servicecertsigner.config.openshift.io/v1alpha1
 kind: APIServiceCABundleInjectorConfig
 caBundleFile: /var/run/configmaps/signing-cabundle/ca-bundle.crt
-authentication:
```
marun: @stlaz Is this just a configuration change that prompts library code to ensure the metrics endpoint is secured? If so, why would that require operator-specific e2e coverage?

stlaz: It would require a library extended test, or at least an integration one, but we would still want to check that the operator is actually behaving the intended way and publishes it's metrics as desired, no?

test/e2e/e2e_test.go
```diff
@@ -292,6 +295,96 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+func GenerateRoute(routeClient *routeclient.RouteV1Client, s *v1.Service, namespace string) (*routev1.Route, error) {
+	labels := make(map[string]string)
+	for k, v := range s.ObjectMeta.Labels {
+		labels[k] = v
+	}
+	route, err := routeClient.Routes(namespace).Create(&routev1.Route{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Route",
+			APIVersion: "route.openshift.io/v1",
+		},
```
stlaz: No need to specify `TypeMeta`, I think the client takes care of it

test/e2e/e2e_test.go
```diff
@@ -292,6 +295,96 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+	service, err := client.CoreV1().Services(namespace).Create(&v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceName,
+			Namespace: namespace,
+			Labels:    map[string]string{"name": deploymentName},
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
```
stlaz: ditto

test/e2e/e2e_test.go
```diff
@@ -292,6 +295,96 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+			Labels:    labels,
+		},
+		Spec: routev1.RouteSpec{
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: s.ObjectMeta.Name,
+			},
+			Port: &routev1.RoutePort{
+				TargetPort: s.Spec.Ports[0].TargetPort,
+			},
```
stlaz: Just use constant names

bindata/v4.0.0/apiservice-cabundle-controller/defaultconfig.yaml
stlaz: I finally got to browse through the code - it was not my point to try to check that the endpoints are authenticated, but that they are actually showing something

marun: Couldn't that be validated in a unit test?

stlaz: I think that that would be a fine general approach to test an endpoint, yes. I wonder how many endpoints we'd like to test this way though.

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,110 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
```
stlaz: Wait for the route to reach "Admitted=True" condition, see `k8s.io/apimachinery/pkg/util/wait` and `PollImmediate()`

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,110 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+
+	httpclient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
+	}
+
+	res, err = httpclient.Get(url)
```
stlaz: Try polling for a bit

bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
```diff
@@ -13,3 +13,19 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - create
```
stlaz: Don't we just need access to one specific configmap? Why does it have all of `get, list, watch, update, create`? Wouldn't get/list just be enough?

bindata/v4.0.0/service-serving-cert-signer-controller/clusterrole.yaml
```diff
@@ -24,3 +24,19 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - create
```
stlaz: ditto

test/e2e/e2e_test.go
```diff
@@ -2,32 +2,37 @@ package e2e
...
 	"github.com/openshift/service-ca-operator/test/util"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
```
stlaz: keep these in a separate block as before

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,140 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
 	})
 }
 
+// GenerateRoute creates a OpenShift routes
+func GenerateRoute(routeClient *routeclient.RouteV1Client, s *v1.Service, namespace string) (*routev1.Route, error) {
+	route, err := routeClient.Routes(namespace).Create(&routev1.Route{
+		TypeMeta: metav1.TypeMeta{},
```
stlaz: Remove this line

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,140 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      s.ObjectMeta.Name,
+			Namespace: s.ObjectMeta.Namespace,
+			Labels:    s.ObjectMeta.Labels,
```
stlaz: I think you can access all the attributes directly from the `s` struct, although generally these attributes of service and route are not logically connected. Good enough for test though.

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,140 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+		},
+		Spec: routev1.RouteSpec{
+			TLS: &routev1.TLSConfig{
+				InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
+				Termination:                   routev1.TLSTerminationPassthrough,
+			},
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: s.ObjectMeta.Name,
```
stlaz: `s.Name`

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,140 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+		return true, nil
+	}); err != nil {
+		return nil, err
+	}
+	return route, nil
+}
+
+func IsAdmitted(route *routev1.Route) bool {
+	for _, ingress := range route.Status.Ingress {
+		if isIngressAdmitted(ingress) {
```
stlaz: merge these two together, only pass the `route.Status.Ingress` in arguments, rename to `isRouteAdmitted()`, we can export this later on if anyone needs it somewhere else.

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,140 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+			admitted = true
+		}
+	}
+	return admitted
+}
+
+// createService creates a service object and returns a pointer pointing to this object on success.
+func createService(client *kubernetes.Clientset, namespace string, serviceName string, deploymentName string) (*v1.Service, error) {
+	service, err := client.CoreV1().Services(namespace).Create(&v1.Service{
+		TypeMeta: metav1.TypeMeta{},
```
stlaz: Remove this line

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,140 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+			res.Body.Close()
+		}
+		client.CoreV1().Services(namespace).Delete(serviceName, &metav1.DeleteOptions{})
+		routeClient.Routes(namespace).Delete(serviceName, &metav1.DeleteOptions{})
+	}()
+	service, err := createService(client, namespace, serviceName, deploymentName)
+	if err != nil {
+		t.Fatalf("could not create a service: %v", err)
+	}
+	route, err := GenerateRoute(routeClient, service, namespace)
```
stlaz: Why the naming and visibility differences of `createService` and `GenerateRoute`?

bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
sohankunkerkar: Yes, it makes sense to me. Initially, I thought of giving all the permissions. Let me correct that.

bindata/v4.0.0/service-serving-cert-signer-controller/clusterrole.yaml
sohankunkerkar: Ack

test/e2e/e2e_test.go
sohankunkerkar: I don't know how to do it. It's automatically rearranged by the `VSCode` editor.

test/e2e/e2e_test.go
sohankunkerkar: Fixed

test/e2e/e2e_test.go
sohankunkerkar: I have added this line after seeing the similar usage in the `createServingCertAnnotatedService` function in this file.

test/e2e/e2e_test.go
sohankunkerkar: Fixed.

test/e2e/e2e_test.go
sohankunkerkar: I can't do that due to below error:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_service-ca-operator/81/pull-ci-openshift-service-ca-operator-master-verify/174

stlaz:
```
diff --git a/test/e2e/e2e_test.go b/test/e2e/e2e_test.go
index 33d327b8..8389cc41 100644
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -14,6 +14,16 @@ import (
        "testing"
        "time"
 
+       v1 "k8s.io/api/core/v1"
+       "k8s.io/apimachinery/pkg/api/errors"
+       metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+       kruntime "k8s.io/apimachinery/pkg/runtime"
+       "k8s.io/apimachinery/pkg/util/intstr"
+       "k8s.io/apimachinery/pkg/util/wait"
+       "k8s.io/client-go/kubernetes"
+       "k8s.io/client-go/rest"
+       "k8s.io/client-go/tools/clientcmd"
+
        routev1 "github.com/openshift/api/route/v1"
        operatorv1client "github.com/openshift/client-go/operator/clientset/versioned"
        routeclient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
@@ -24,15 +34,6 @@ import (
        "github.com/openshift/service-ca-operator/pkg/operator"
        "github.com/openshift/service-ca-operator/pkg/operator/operatorclient"
        "github.com/openshift/service-ca-operator/test/util"
-       v1 "k8s.io/api/core/v1"
-       "k8s.io/apimachinery/pkg/api/errors"
-       metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-       kruntime "k8s.io/apimachinery/pkg/runtime"
-       "k8s.io/apimachinery/pkg/util/intstr"
-       "k8s.io/apimachinery/pkg/util/wait"
-       "k8s.io/client-go/kubernetes"
-       "k8s.io/client-go/rest"
-       "k8s.io/client-go/tools/clientcmd"
 )
 
 const (

```
works just fine for me even in VS Code.

test/e2e/e2e_test.go
```diff
@@ -292,6 +297,132 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+
+func IsRouteAdmitted(route *routev1.Route) bool {
+	admitted := false
+	for _, ingress := range route.Status.Ingress {
+		for _, condition := range ingress.Conditions {
+			if condition.Type == routev1.RouteAdmitted && condition.Status == v1.ConditionTrue {
+				admitted = true
+			}
+		}
+		return admitted
```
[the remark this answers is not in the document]
sohankunkerkar: fixed

test/e2e/e2e_test.go
sohankunkerkar: Fixed

test/e2e/e2e_test.go
```diff
@@ -292,6 +298,131 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+	}
+	return route, nil
+}
+
+func IsRouteAdmitted(route *routev1.Route) bool {
+	admitted := false
+	for _, ingress := range route.Status.Ingress {
+		for _, condition := range ingress.Conditions {
+			if condition.Type == routev1.RouteAdmitted && condition.Status == v1.ConditionTrue {
+				admitted = true
```
stlaz: this now loops even though it long knows `admitted = true`

sohankunkerkar: yeah, I will add a `break` statment after `admitted = true`

stlaz: well, that will then just make this loop in the wrapping for loop, you can `return true` right away

test/e2e/e2e_test.go
```diff
@@ -579,7 +709,25 @@ func TestE2E(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error getting admin client: %v", err)
 	}
+	routeClient, err := routeclient.NewForConfig(adminConfig)
+	if err != nil {
+		t.Fatalf("error getting route client: %v", err)
+	}
+
+	// Test case to validate if metrics endpoint for the apiservice-cabundle-injector controller is secured.
```
stlaz: `for _, controller := []string{"apiservice-cabundle-injector", "configmap-cabundle-injector", "service-serving-cert-signer"}`, derive your service name for `testMetricsEndpoint()` from the controller name, last argument of that function is constant.

Improve the test case name, at leat to `<controller-name> /metrics`

bindata/v4.0.0/apiservice-cabundle-controller/defaultconfig.yaml
marun: I've identified 10 operators in the service ca compatibility audit that should be tested this way, and there are likely more.

test/e2e/e2e_test.go
```diff
@@ -292,6 +299,134 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
 	})
 }
 
+// createRouteForService creates a OpenShift routes
+func createRouteForService(routeClient *routeclient.RouteV1Client, s *v1.Service, namespace string) (*routev1.Route, error) {
```
marun: Exposing the metrics endpoint for a route should not be necessary here - way too complicated. The endpoint should be checked from inside the cluster

bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
```diff
@@ -13,3 +13,16 @@ rules:
   - watch
   - update
   - patch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
```
marun: Is it necessary for the controller to access configmaps across the entire cluster, or only in the `service-ca namespace`?

test/e2e/e2e_test.go
```diff
@@ -579,6 +714,17 @@ func TestE2E(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error getting admin client: %v", err)
 	}
+	routeClient, err := routeclient.NewForConfig(adminConfig)
+	if err != nil {
+		t.Fatalf("error getting route client: %v", err)
+	}
+
+	// Test case to validate if metrics endpoint for the controllers are secured.
+	for _, controller := range []string{"apiservice-cabundle-injector", "configmap-cabundle-injector", "service-serving-cert-signer"} {
```
marun: What about the operator?

test/e2e/e2e_test.go
```diff
@@ -292,6 +299,134 @@ func pollForConfigMapChange(client *kubernetes.Clientset, compareConfigMap *v1.C
...
+	if err != nil {
+		t.Fatalf("could not create a route: %v", err)
+	}
+
+	url := fmt.Sprintf("https://%s/metrics", route.Spec.Host)
+
+	httpclient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
```
marun: - This needs to be false. 
- A metrics endpoint needs to be secured with a service ca-provided serving cert. 
  - The metrics endpoint is automatically secured by a serving cert provided the pod template has the serving cert secret mounted to `/var/run/secrets/serving-cert` (this is already true for the operator).
  - The controller framework will generate an insecure self-signed certificate on startup and will exit the pod when the cert files are updated, so even the service ca controllers and operators can be secured with serving certs.  
  - Configuring each controller with a service ca-provided ca bundle will require creating an annotated service ( `service.beta.openshift.io/serving-cert-secret-name`) that exposes the metrics endpoint.
- The client to access the metrics endpoint needs to be configured with a service ca-provided ca bundle (see the rotation tests for how to accomplish this).

@stlaz Who is going to watch these new endpoints, and how do they discover them?

bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
sohankunkerkar: I have added that change after seeing the following error:
```
configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:openshift-service-ca:apiservice-cabundle-injector-sa" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```

test/e2e/e2e_test.go
sohankunkerkar: The test cases are added in conjunction with the changes done in this PR. I can add another test case if that serves the purpose.

bindata/v4.0.0/apiservice-cabundle-controller/clusterrole.yaml
marun: That suggests limiting the scope of authorization to that resource in that namespace rather than granting global access to configmaps.

test/e2e/e2e_test.go
sohankunkerkar: I think that sounds reasonable to me. I have designed this logic after having some discussion with @stlaz 
@stlaz wdyt?

test/e2e/e2e_test.go
marun: A metrics endpoint is a metrics endpoint - it is exposed by the controller framework that both the operator and its controllers use. A separate test should not be required.

bindata/v4.0.0/apiservice-cabundle-controller/defaultconfig.yaml
sohankunkerkar: I think what @marun suggested makes sense to me.

test/e2e/e2e_test.go
stlaz: either is fine I think, I noticed people from console doing this and it looked like a good idea to me, for testing

test/e2e/e2e_test.go
stlaz: Those endpoints are not new, they are being scraped by monitoring operator, they most probably use their service account for that.

The InsecureSkipVerify is here as you're not trying to see whether the endpoint is secured by a valid certificate, that it should always be, otherwise scraping the data would not work, it merely tested that you can't scrape it unauthenticated.

marun: The use of `InsecureSkipVerify` is masking a bug.  Metrics endpoints are not secured automatically by a valid certificate, as per my previous comment, and anyone trying to connect to the endpoint enabled by the current state of this PR is going to be unable to validate the cert.

Can you please point me to how the monitoring operator knows how to read from these endpoints? Given that the endpoint cert is not secure in the current iteration of this PR, I am interested to know how to verify on the monitoring operator side whether the endpoints it is charged with collecting from are configured properly.

test/e2e/e2e_test.go
marun: A common pattern in upstream kube when needing to validate a cluster-internal endpoint is to launch a busybox pod that connects to the endpoint. This pattern limits the scope of test dependency. Why is it desirable for validation of internal connectivity to depend on external connectivity?

marun:
Given that these metrics endpoints are not currently being used by the cluster, I think there are 2 potential paths forward:

 - Disable the endpoints 
    - Who is intended to use them in their current form? 
      - They are using insecure self-signed certificates.
 - Configure prometheus monitoring properly
   - According to the [monitoring doc](https://docs.google.com/document/d/1rCKAYTrYMESjJDyJ0KvNap05NNukmNXVVi6MShISnOw/edit) this is the responsibility of all operators anyway.
     - The doc suggests that properly instrumenting an operator is likely to take a few hours.

I do not believe that enabling authn/authz on the metrics endpoints and writing e2e tests to verify that configuration, as currently proposed, is worth pursuing. Testing what is essentially a function of library go (configuration of metrics endpoints) should be out-of-scope for this operator.  If the endpoints are intended to be used, testing that prometheus monitoring is properly configured to use them would ensure proper configuration without requiring bespoke testing. There is already at least [one example](https://github.com/openshift/origin/pull/24169) of testing that prometheus is receiving alerts and this pattern could probably be generalized for inclusion in library-go (cc: @sallyom).

marun:
LGTM

@stlaz I asked for the e2e test be removed. Ensuring that the generic operator config doesn't disable the auth is sufficient to ensure that the endpoint isn't accessible to unauthorized access. A follow-on PR can configure prometheus to read from the endpoint and leverage a generic test for prometheus metric configuration that @sallyom and @sanchezl are working on adding to library-go.

stlaz:
/lgtm
@marun sounds good to me

PR #83, opened 2019-11-14
Revert the removal of the service CA from serving certs

marun:
The addition of CA rotation also saw the removal of the signing CA from generated serving certificates. This broke legacy clients that depended on this behavior. To support code-free changes to those legacy clients, this change maintains the old behavior of including the signing CA so long as the alpha annotation is used. Beta and above will not include the signing CA.

### DIFF
--- a/pkg/controller/servingcert/controller/secret_creating_controller.go
+++ b/pkg/controller/servingcert/controller/secret_creating_controller.go
@@ -333,17 +333,14 @@ func MakeServingCert(dnsSuffix string, ca *crypto.CA, intermediateCACert *x509.C
 	if err != nil {
 		return nil, err
 	}
-	if len(servingCert.Certs) > 1 {
-		// Only include the generated cert. The issuing cert will be provided to
-		// clients via the ca bundle.
-		servingCert.Certs = servingCert.Certs[:1]
-	}
+
 	// Including the intermediate cert will ensure that clients with a
 	// stale ca bundle (containing the previous CA but not the current
 	// one) will be able to trust the serving cert.
 	if intermediateCACert != nil {
 		servingCert.Certs = append(servingCert.Certs, intermediateCACert)
 	}
+
 	return servingCert, nil
 }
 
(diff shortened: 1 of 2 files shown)


marun:
/hold

This PR is on hold pending the approval of a proposed change to the rotation enhancement:

https://github.com/openshift/enhancements/pull/109

marun:
/hold cancel

This PR has been updated to revert to the previous behavior of including the issuing CA cert with serving certs.

pkg/controller/servingcert/controller/secret_creating_controller.go
```diff
@@ -333,17 +333,14 @@ func MakeServingCert(dnsSuffix string, ca *crypto.CA, intermediateCACert *x509.C
 	if err != nil {
 		return nil, err
 	}
-	if len(servingCert.Certs) > 1 {
-		// Only include the generated cert. The issuing cert will be provided to
-		// clients via the ca bundle.
-		servingCert.Certs = servingCert.Certs[:1]
-	}
+
```
stlaz: How about we keep the old behavior for `AlphaServingCertSecretAnnotation` (I think that was your original proposal), have the newer one for `beta` version of the annotation, then claim the `alpha` annotation as deprecated for 4.4 and remove it completely in 4.5? That should effectively move all the users who misuse this bundle to do the right thing in 4.5.

marun: That was what this PR originally proposed, but @sttts convinced me that changing the behavior of the beta annotation constituted a breaking change with a potential customer impact. I don't think the current behavior justifies a breaking change. As per the comment I added to the api file, I think it is preferable to fix the behavior for a GA annotation rather than introducing a breaking change.

stlaz: Would it make sense to make betav2 annotation with the more correct behavior, then? I don't think we would like to leap as much as doing a breaking change from beta to GA, in case we missed something else.

marun: I don't think there's value in worrying about fixing this behavior before GA.  And anyone using the serving cert as a ca bundle will not get the benefit of trust-maintaining rotation, so that ill-advised usage is likely to end even without a GA annotation.

stlaz: /lgtm
Fair enough for me

PR #85, opened 2019-11-15
Support setting CA validity duration via unsupportedConfigOverrides

marun:
Setting the duration of CA validity allows setting a short duration to test operator compatibility with CA rotation.

This PR is required by the proposed periodic rotation compatibility job:  https://github.com/openshift/release/pull/5930

### DIFF
--- a/pkg/operator/cert.go
+++ b/pkg/operator/cert.go
@@ -1,0 +1,58 @@
+package operator
+
+import (
+	"crypto/x509"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/service-ca-operator/pkg/operator/util"
+)
+
+// maybeUpdateExpiry updates the expiry of the self-signed CA and returns it if the
+// provided duration is greater than zero. If an error occurs or the duration is
+// zero, the original CA is returned.
+func maybeUpdateExpiry(caConfig *crypto.TLSCertificateConfig, duration time.Duration) (*crypto.TLSCertificateConfig, error) {
+	if duration.Nanoseconds() == 0 {
+		return caConfig, nil
+	}
+	// operator-go's cert generation only accepts an expiry in days. To
+	// ensure support for all validity durations, even those less than a
+	// day, renew the self-signed cert.
+	renewedCAConfig, err := RenewSelfSignedCertificate(caConfig, duration, false)
+	if err != nil {
+		return caConfig, err
+	}
+	return renewedCAConfig, nil
+}
+
+// RenewSelfSignedCertificate updates the expiry and optionally the serial number of the
+// provided self-signed CA.
+func RenewSelfSignedCertificate(caConfig *crypto.TLSCertificateConfig, duration time.Duration, incrementSerial bool) (*crypto.TLSCertificateConfig, error) {
+	caCert := caConfig.Certs[0]
+
+	// Copy the ca cert to avoid modifying the one provided
+	template, err := x509.ParseCertificate(caCert.Raw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to copy ca certificate: %v", err)
+	}
+
+	// Update the expiry
+	expiry := time.Now().Add(duration)
+	template.NotAfter = expiry
+
+	if incrementSerial {
+		template.SerialNumber = template.SerialNumber.Add(template.SerialNumber, big.NewInt(1))
+	}
+
+	renewedCACert, err := util.CreateCertificate(template, template, caCert.PublicKey, caConfig.Key)
+	if err != nil {
+		return nil, fmt.Errorf("error creating ca certificate: %v", err)
+	}
+
+	return &crypto.TLSCertificateConfig{
+		Certs: []*x509.Certificate{renewedCACert},
+		Key:   caConfig.Key,
+	}, nil
+}
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -2,19 +2,39 @@ package operator
 
 import (
 	"encoding/json"
+	"time"
 )
 
 type unsupportedServiceCAConfig struct {
+	CAConfig caConfig `json:"caConfig"`
+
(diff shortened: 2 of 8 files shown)


marun:
Rebased and ready for review

pkg/operator/rotate.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	signingCertificateLifetimeInDays = 365 // 1 year
+	SigningCertificateLifetimeInDays = 365 // 1 year
```
stlaz: If someone in the future changes this to `var`, I will eat them

pkg/operator/cert.go
```diff
@@ -0,0 +1,58 @@
...
+
+// maybeUpdateExpiry updates the expiry of the self-signed CA and returns it if the
+// provided duration is greater than zero. If an error occurs or the duration is
+// zero, the original CA is returned.
+func maybeUpdateExpiry(caConfig *crypto.TLSCertificateConfig, duration time.Duration) (*crypto.TLSCertificateConfig, error) {
+	if duration.Nanoseconds() == 0 {
+		return caConfig, nil
+	}
+	// operator-go's cert generation only accepts an expiry in days. To
+	// ensure support for all validity durations, even those less than a
```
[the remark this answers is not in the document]
marun: Having a flexible initial expiry is required to support validating that cert expiry is handled correctly in the context of a periodic job that takes less than a day to execute (see the description of the [proposed rotation job](https://docs.google.com/document/d/1NB2wUf9e8XScfVM6jFBl8VuLYG6-3uV63eUpqmYE8Ts/edit#heading=h.8rgsj08xt5tp)). Exploratory testing has a similar requirement. If qa or someone else wants to see what happens when a cert expires, I think it makes sense to allow the specification of the appropriate interval (likely in minutes or hours) rather than forcing them to wait a whole day or depend on clock manipulation.

So, there is a requirement. And if this repo is the only one that requires it, I'm not sure why that suggests the effort of updating library-go (especially for an unsupported api).

To your point, though, it is not great UX for a general-purpose option to set the CA expiry in at most hours (since that's the maximum granularity that `Duration` supports). I think that suggests limiting the scope of the option to the initial cert (or manual rotations, since that's the same code path) and changing the name to `ValidityDurationForTesting` to make its use explicit. Forced and automatic rotations would use the default and if there is user demand for varying that duration in the future that would be covered by another option (e.g. `GracefulValidityDuration`) specified in days. WDYT?

pkg/operator/cert.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/service-ca-operator/pkg/operator/util"
+)
+
+// maybeUpdateExpiry updates the expiry of the self-signed CA and returns it if the
+// provided duration is greater than zero. If an error occurs or the duration is
+// zero, the original CA is returned.
+func maybeUpdateExpiry(caConfig *crypto.TLSCertificateConfig, duration time.Duration) (*crypto.TLSCertificateConfig, error) {
+	if duration.Nanoseconds() == 0 {
```
sttts: isn’t there duration.IsZero?

pkg/operator/config.go
```diff
@@ -2,19 +2,37 @@ package operator
...
+	// +optional
 	Enabled bool `json:"enabled"`
 }
 
 type forceRotationConfig struct {
+	// reason indicates why a rotation of the signing CA should be forced. If the
+	// reason is not empty and has not been recorded as an annotation on the signing
+	// secret, the rotation of the signing CA will be triggered at most once.
+	// +optional
 	Reason string `json:"reason"`
```
sttts: API lgtm

pkg/operator/rotate.go
```diff
@@ -108,6 +103,12 @@ func maybeRotateSigningSecret(secret *corev1.Secret, currentCACert *x509.Certifi
 		return "", err
 	}
 
+	// Set a custom expiry if one was provided
+	signingCA.config, err = maybeUpdateExpiry(signingCA.config, serviceCAConfig.CAConfig.ValidityDuration)
+	if err != nil {
+		return "", fmt.Errorf("Error renewing ca for custom duration: %v", err)
```
sttts: lower case errors

pkg/operator/cert.go
marun: Doh.  Done.

pkg/operator/rotate.go
marun: Done.

pkg/operator/config.go
marun: So you do not share @stlaz's concerns regarding `ValidityDuration` being specified in at most hours?

marun:
/hold cancel

Updated a couple of error messages in latest push, nothing functional.

pkg/operator/config.go
sttts: Probably >= 1h makes sense. So either document the minimum or switch to hours as integer.

sttts:
Sgtm overall (would switch to hours).

pkg/operator/config.go
marun: What would be the value in switching to an integer representing hours given that `metav1.Duration` supports providing a time in hours (e.g. `1h`) and supporting a duration in minutes (e.g. `20m`) is potentially useful for testing?

sttts: See @liggitt's talk at contributor summit: `metav1.Duration` is considered a failure.

marun: I don't see the video up yet, but I'm assuming you're suggesting that metav1.Duration not be used? That reinforces my suggestion for `ValidityDurationForTesting` and denominating it in minutes.  If `ValidityDuration` becomes a thing to set the CA expiry for production use, it would likely be in days.

sttts: ValidityDurationForTesting is fine with me as well. Let's decide and move forward.

marun: I've update the name to `ValidityDurationForTesting` and left it as `metav1.Duration`, hopefully that makes sense for you.

marun:
Rebased

PR #87, opened 2019-11-28
Update to go modules

marun:


go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/atomic v1.4.0 // indirect
+	go.uber.org/multierr v1.1.0 // indirect
+	go.uber.org/zap v1.10.0 // indirect
+	k8s.io/api v0.0.0-20190918155943-95b840bb6a1f
+	k8s.io/apiextensions-apiserver v0.0.0-20190918161926-8f644eb6e783 // indirect
+	k8s.io/apimachinery v0.0.0-20190913080033-27d36303b655
+	k8s.io/client-go v0.0.0-20190918160344-1fbdaa4c8d90
+	k8s.io/component-base v0.0.0-20190918160511-547f6c5d7090
+	k8s.io/klog v0.4.0
+	k8s.io/kube-aggregator v0.0.0-20190918161219-8c8f079fddc3
```
sanchezl: Use `replace` directives to "pin" the k8s versions.

The [goodmod tool](https://github.com/mfojtik/goodmod) can help with this.

marun: Is this documented somewhere?

I've updated quite a few repos using gomod before and it's generally sufficient to `go get` the appropriate revision of dependencies that need to be updated (e.g. `go get `k8s.io/kube-aggregator@<version>`).

In the case of the versions that are specified here, they are identical to the versions used by glide.

I'm sure there's good reason for using something like goodmod for repos where vendoring is complicated, but I'm loath to have to require it for simple repos if I already understand how to use go mod.

cc: @mfojtik

PR #88, opened 2019-12-11
Remove unnecessary use of API-based controller configuration

marun:
The configuration previously supplied via API to controllers managed by this operator was entirely dependent on the configuration of their deployments.


marun:
Updated to ensure removal of the now unused configmaps.

marun:
Test failures are legit, marking WIP or now.

sttts:
Sgtm.

pkg/operator/sync.go
```diff
@@ -6,6 +6,8 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
+var cleanupDeprecatedDone bool
```
sttts: nit: I know we set workers=1 when creating controllers. But technically this bool needs to be protected.

marun: Done

pkg/operator/sync.go
```diff
@@ -1,11 +1,18 @@
...
 import (
+	"sync"
+
 	"k8s.io/klog"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
+var cleanupDeprecated = struct {
+	sync.RWMutex
```
sttts: sync.Mutex

pkg/operator/sync.go
```diff
@@ -14,6 +21,17 @@ func syncControllers(c serviceCAOperator, operatorConfig *operatorv1.ServiceCA)
...
+	cleanupDeprecated.Lock()
+	if !cleanupDeprecated.done {
+		err := cleanupDeprecatedResources(c)
+		if err != nil {
+			cleanupDeprecated.Unlock()
+			return err
+		}
+		cleanupDeprecated.done = true
+	}
+	cleanupDeprecated.Unlock()
```
sttts: alternative, less ugly: get an configmap informer for the operator namespace and use its lister to look for the resource unconditionally.

marun: Less ugly for this PR maybe, but that seems like huge overhead given the cleanup required by the controller unification PR:

https://github.com/openshift/service-ca-operator/pull/89/commits/d41d2c6e31f60d9a628adfaae4b5b79c66381d76#diff-bd2d316d0e70d82715b2108d2a21adb8R235

pkg/operator/sync_common.go
```diff
@@ -262,3 +230,23 @@ func manageDeployment(client appsclientv1.AppsV1Interface, eventRecorder events.
 func serviceServingCertSignerName() string {
 	return fmt.Sprintf("%s@%d", "openshift-service-serving-signer", time.Now().Unix())
 }
+
+// cleanupDeprecatedResources ensures the deletion of resources no longer required by the operator.
+func cleanupDeprecatedResources(c serviceCAOperator) error {
```
sttts: while fine here, this could become a cleanup helper for library-go using the dynamic client. Wdyt?

marun: Do you mean allowing a simple way to declare resources that are to be removed, as follows?

```golang
type DeletionTarget { 
  Object pkgruntime.Object
  Name string
  Namespace string
}

targets := []DeletionTargets{
  {v1.ClusterRole{}, "foo"},
  {myType{}, "bar", "foo"},
}
```

This should be sufficient input to a helper using the dynamic or controller-runtime client.

Or did you want to avoid typing altogether? As much as I'm happy dealing with GVK or GVR, I'm not sure that's nice UX for most.

sttts: we have `Coordinate`s for arbitrary resource I think.

marun: I see `ResourceLocation`, which can be used for name+namespace, is that what you meant? Is there something similar to describe type details? I'm not sure how it could be easier than an obj or gvr, though, given that there is inherent complexity in construing the rest url.

PR #89, opened 2019-12-12
Deploy all controllers as one process

marun:
Previously the 2 injection controllers and the serving cert controller were deployed separately. Deploying them as a single process simplifies maintenance and the proposed configuration of metrics.


marun:
@sttts @stlaz Ready for review

bindata/v4.0.0/service-serving-cert-signer-controller/ns.yaml
```diff
@@ -1,8 +0,0 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: openshift-service-ca
-  labels:
-    openshift.io/run-level: "1"
-  annotations:
-    openshift.io/node-selector: ""
```
sttts: where is this gone?

bindata/v4.0.0/controller/clusterrole.yaml
```diff
@@ -24,13 +24,22 @@ rules:
...
   - ""
-  resourceNames:
-  - extension-apiserver-authentication
   resources:
   - configmaps
   verbs:
   - get
   - list
   - watch
+  - update
```
sttts: no create?

pkg/operator/starter.go
```diff
@@ -29,11 +29,7 @@ const (
 	operatorVersionEnvName = "OPERATOR_IMAGE_VERSION"
 )
 
-var targetDeploymentNames = sets.NewString(
-	api.SignerControllerDeploymentName,
-	api.APIServiceInjectorDeploymentName,
-	api.ConfigMapInjectorDeploymentName,
-)
+var targetDeploymentNames = sets.NewString(api.ServiceCADeploymentName)
```
sttts: that's a big set.

bindata/v4.0.0/controller/clusterrole.yaml
stlaz: it only injects the cert IIRC, create should not be needed

bindata/v4.0.0/service-serving-cert-signer-controller/ns.yaml
marun: All 3 controllers had the same content for `ns.yaml` and even though I renamed the cert signing path git chose to use the one from the apiserver injector.

pkg/cmd/controller/cmd.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/openshift/service-ca-operator/pkg/controller"
+	"github.com/openshift/service-ca-operator/pkg/version"
+)
+
+func NewController() *cobra.Command {
+	cmd := controllercmd.
+		NewControllerCommandConfig("service-ca-controller", version.Get(), controller.StartServiceCAControllers).
+		NewCommand()
+	cmd.Use = "controller"
```
stlaz: Nit: `ca-injectors`? `controller` is just too vague

marun: It's not just ca injectors.  It's the cert secret generating controller too. The new normal for this operator is an 'operator' command and the 'controller' command that it manages.

pkg/operator/starter.go
marun: Room for cleanup around status, for sure, but I'd rather tackle that separately. The main intent of this PR is reducing the resource footprint of the operator in advance of the bundle injector refactor to avoid unnecessary churn in resources managed by the operator.

bindata/v4.0.0/controller/clusterrole.yaml
marun: The tests are passing, and configmap injection is well-tested.

pkg/cmd/controller/cmd.go
stlaz: My bad, I've gone through the PR a bit too fast. Name's fine, then.

marun:
Beyond rebase, this PR needs to be updated to support downgrade. Proposed solution is to exit the new unified controller if the old controller deployments are present.

marun:
/hold cancel

@deads2k I've added a [simple polling check](https://github.com/openshift/service-ca-operator/pull/89/files#diff-8248cb6ce5099ca35413cfe7a8b16941R25) for a downgrade, PTAL. I thought it overkill to use a full informer-based controller given how infrequent the checks need to be, and though it preferable for the pod to scale its managing deployment down rather than simply exiting.

pkg/controller/starter.go
```diff
@@ -0,0 +1,81 @@
...
+
+	<-ctx.Done()
+
+	return fmt.Errorf("stopped")
+}
+
+// maybeScaleDown scales down the deployment managing this pod if a downgrade has
+// occurred to a legacy version of the operator that uses a deployment per
+// controller.
+func maybeScaleDown(deployClient appsv1.DeploymentInterface) {
```
stlaz: nit: can you add a note about when all of this is going to be safe to be removed?

marun: Done.

marun: I've added a note at the call point indicating when removal is an option.

stlaz: I meant an exact version - it should be safe for it to be removed in 4.5 since downgrade would still have to pass 4.4, no?

marun: Would it possible to upgrade from 4.2 to 4.5?  Would it be possible then to downgrade from 4.5 to 4.2? 

Also, can you point me to documentation on how upgrades work? I don't think I understand what upgrade strategies are supported for openshift.

marun:
Marking WIP in response to a conversation on slack that suggested moving the downgrade handling to 4.3. I'll be removing the downgrade handling from this PR.

marun:
@stlaz Removed the downgrade handling in favor of updating 4.3 with support for cleaning up the new resources. PTAL

marun:
/hold 

Need to update cleanup of the 4.3 deployment to be periodic rather than one-shot (as per @deads2k).  It should be safe to continue to clean up the non-deployment resources less often, since they won't have an operational impact.

marun:
/hold cancel 

Updated to poll for removal of 4.3 deployment and added comments indicating that removal of 4.3 cleanup can be done in 4.5 (as per @stlaz).

bindata/v4.0.0/controller/role.yaml
```diff
@@ -37,3 +37,5 @@ rules:
   - get
   - list
   - watch
+  # Support scaling the unified deployment to zero on downgrade to multiple independent deployments
+  - update
```
stlaz: This should be `delete` now

marun: Good catch, and prompt to ensure deletion of the other resources (and indicate that this permission can be revoked in 4.5).

marun: Updated, PTAL

marun:
Rebased

pkg/operator/starter.go
```diff
@@ -117,6 +116,28 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	//
+	// This code can be removed in 4.5 since downgrade to 4.3 will no longer be possible.
+	deployClient := kubeClient.AppsV1().Deployments(operatorclient.TargetNamespace)
+	go wait.Until(func() {
+		// List rather than blindly deleting so that there are max 1 request/minute
+		// instead of 3 (one for each 4.3 deployment).
+		deploys, err := deployClient.List(metav1.ListOptions{})
+		if err != nil {
+			klog.Warningf("Failed to list deployments when searching for 4.3 deployments to remove: %v", err)
+		}
```
sttts: return.  `deploys` could be nil leading to a crash in the next line.

marun: Yikes, good catch. Apparently I'm nothing without tests!

PR #90, opened 2019-12-19
Configure metrics collection for operator

marun:
Previously standard controller metrics for the operator were being exposed but not scraped by prometheus. This change adds the necessary resources and configuration for metrics collection by prometheus.

Enabling metrics collection for the controllers managed by the operator will be done after #89 to avoid adding resources prior to controller unification.


marun:
@stlaz Ready for review!

PR #91, opened 2020-01-03
bump builder image to golang-1.13 in dockerfiles

stlaz:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS builder
```
tnozicka: I'd drop this whole file as we have done in other repos. There is no OKD yet and `registry.svc.ci.openshift.org/openshift/origin-v4.0:base` is actually OCP anyways

marun:
@stlaz I'm assuming go.mod should be updated to 1.13 too?

https://github.com/openshift/service-ca-operator/blob/master/go.mod#L3

stlaz:
@tnozicka apparently I can't simply remove the file since it's used in the build, but I should be able to do that once https://github.com/openshift/release/pull/6549 merges AFAIU

tnozicka:
dockerfile FROM changes need to go in first, so you need to bump it there to 1.13 too and delete in a follow up. Or split the release PR into 2 (first to fix just the dockerfile location)

PR #92, opened 2020-01-07
remove dockerfile and add depency checking targets

stlaz:
/assign @marun 

/cc @tnozicka


marun:
/lgtm

Nice to see explicit version tags supported for go mod (vs the opaque computed tags)!

PR #93, opened 2020-01-14
Bug 1791022: Ensure removal of 4.4-only resources on downgrade to 4.3

marun:
This PR is in support of a 4.4 PR that unifies all service ca controllers into a single deployment to simplify maintenance and administration: https://github.com/openshift/service-ca-operator/pull/89

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -4,9 +4,13 @@ import (
 	"fmt"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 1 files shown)


marun:
@sttts @stlaz This PR would ideally merge in time for the next 4.3.z release.

pkg/operator/starter.go
```diff
@@ -114,6 +118,24 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+	deployName := "service-ca"
+	go wait.Until(func() {
+		_, err := deployClient.Get(deployName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return
+		}
+		if err != nil {
+			klog.Warningf("Failed to retrieve 4.4 deployment: %v", err)
+		}
+		err = deployClient.Delete(deployName, &metav1.DeleteOptions{})
```
sttts: why not delete directly, why a GET?

Shouldn't we stop the loop as soon as it succeeds?

marun: Done.

pkg/operator/sync.go
```diff
@@ -14,6 +36,16 @@ func syncControllers(c serviceCAOperator, operatorConfig *operatorv1.ServiceCA)
 		return err
 	}
 
+	// Remove resources related to the 4.4 controller deployment at most once. These
+	// resources don't constitute an operational concern, so it is not necessarily to
+	// monitor for their presence.
+	//
+	// The 4.4 deployment does have an operational impact, and is continually monitored
+	// for removal via a goroutine started in RunOperator.
+	if err := once.Do(func() error { return cleanupUnifiedDeployment(c) }); err != nil {
```
sttts: why is there no loop around these like it is in the other case?

marun: The enclosing is the operator sync loop.

marun:
/hold

Need to revert the deployment cleanup to be continuous rather than once as per @deads2k

marun:
/hold cancel

Updated to be in keeping with https://github.com/openshift/cluster-openshift-apiserver-operator/pull/313, PTAL.

pkg/operator/sync.go
```diff
@@ -14,6 +36,16 @@ func syncControllers(c serviceCAOperator, operatorConfig *operatorv1.ServiceCA)
 		return err
 	}
 
+	// Remove resources related to the 4.4 controller deployment at most once. These
```
marun: @deads2k Is it acceptable to remove the resources related to the deployment at most once rather than continually, since the resources in question have no operational impact without the deployment (which is continually removed by the starter)?

pkg/operator/starter.go
stlaz: I think the original code is still in place, did you add your changes to your latest commit?

sttts: Get is still there. But I don't care too much.

sttts: But I would prefer to see the loop stop when the deployment is gone.

marun: I reverted the changes as per @deads2k. iirc his assertion is that continual reconciliation is required to avoid confusing users when behavior is inconsistent (removes a deployment on startup not subsequently).

As to why get, I thought it would be preferable to see 'get' in the log vs 'delete' if continually reconciling but maybe that's specious logic?

sttts: ok, fine with me.

sttts: Am still confused why the deployment is different from the other resources in `cleanupUnifiedDeployment`.

marun: That's what I asked in https://github.com/openshift/service-ca-operator/pull/93#discussion_r382408155. Those other resources don't have an operational impact for the operator, so deleting them isn't so critical, but the principle of consistent behavior suggests continual reconciliation for them too. 

I guess I've answered my own question. Maybe reconcile on a longer timeline - 20m?

pkg/operator/sync.go
sttts: I don't care too much. It's good enough to do it once.

pkg/operator/sync.go
sttts: The combination with `sync.Once` confused me. Anyway, seems to be good enough.

pkg/operator/starter.go
sttts: Fair enough. 20m is also fine, so is 1m.

marun:
/hold cancel

Updated to attempt removal of 4.4 resources every 20 minutes.

PR #95, opened 2020-01-15
Bug 1803282: enable signer ca expiry timer metrics

vareti:
This PR is an attempt to add expiry time and time left metrics before the signer CA expires to Prometheus collector

### DIFF
--- a/pkg/operator/metrics/metrics.go
+++ b/pkg/operator/metrics/metrics.go
@@ -1,0 +1,21 @@
+package metrics
+
+import (
+	"time"
+
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+var expiryTime = metrics.NewGauge(&metrics.GaugeOpts{
+	Name: "service_ca_expiry_time_seconds",
+	Help: "Reports the expiry time of service CA used in the cluster.",
+})
+
+func init() {
+	legacyregistry.MustRegister(expiryTime)
+}
+
+func SetCAExpiry(notAfter time.Time) {
+	expiryTime.Set(float64(notAfter.Unix()))
+}
--- a/pkg/operator/sync_common.go
+++ b/pkg/operator/sync_common.go
@@ -116,6 +116,7 @@ func manageSignerCA(client coreclientv1.SecretsGetter, eventRecorder events.Reco
 			return false, fmt.Errorf("failed to rotate signing CA: %v", err)
 		}
 		if len(rotationMsg) == 0 {
+			metrics.SetCAExpiry(existingCert.NotAfter)
 			return false, nil
 		}
 		// Ensure the updated existing secret is applied below
--- a/pkg/operator/sync_common.go
+++ b/pkg/operator/sync_common.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/x509"
 	"fmt"
+	"github.com/openshift/service-ca-operator/pkg/operator/metrics"
 	"os"
 	"time"
 
@@ -120,6 +121,12 @@ func manageSignerCA(client coreclientv1.SecretsGetter, eventRecorder events.Reco
 		// Ensure the updated existing secret is applied below
 		secret = existing
 	}
+
+	certs, err := cert.ParseCertsPEM(secret.Data[corev1.TLSCertKey])
+	if err != nil {
+		return false, err
+	}
+	metrics.SetCAExpiry(certs[0].NotAfter)
 
 	_, mod, err := resourceapply.ApplySecret(client, eventRecorder, secret)
 
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -686,7 +686,7 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
 func runPromQueryForVector(t *testing.T, promClient prometheusv1.API, query string, sampleTime time.Time) (model.Vector, error) {
 	results, warnings, err := promClient.Query(context.Background(), query, sampleTime)
 	if err != nil {
-		return model.Vector{}, err
+		return nil, err
 	}
 	if len(warnings) > 0 {
 		t.Logf("prometheus query emitted warnings: %v", warnings)
@@ -694,7 +694,7 @@ func runPromQueryForVector(t *testing.T, promClient prometheusv1.API, query stri
 
 	result, ok := results.(model.Vector)
 	if !ok {
-		return model.Vector{}, fmt.Errorf("expecting vector type result, found: %v ", reflect.TypeOf(results))
+		return nil, fmt.Errorf("expecting vector type result, found: %v ", reflect.TypeOf(results))
 	}
 
 	return result, nil
@@ -735,7 +735,7 @@ func checkServiceCAMetrics(t *testing.T, client *kubernetes.Clientset, promClien
 		}
 
 		if float64(want.Unix()) != float64(rawExpiryTime.Value) {
-			t.Fatalf("service ca expiry time mismatch expected %v observed %v", float64(want.UnixNano()), float64(rawExpiryTime.Value))
+			t.Fatalf("service ca expiry time mismatch expected %v observed %v", float64(want.Unix()), float64(rawExpiryTime.Value))
 		}
 
 		return true, nil
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
@@ -662,18 +663,10 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
 
 	err := wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
 		query := fmt.Sprintf("up{namespace=\"%s\"}", namespace)
-		rawResult, warnings, err := promClient.Query(context.Background(), query, time.Now())
+		resultVector, err := runPromQueryForVector(t, promClient, query, time.Now())
 		if err != nil {
 			t.Errorf("failed to execute prometheus query: %v", err)
 			return false, nil
-		}
-		if len(warnings) > 0 {
-			t.Logf("prometheus query emitted warnings: %v", warnings)
-		}
-
-		resultVector, ok := rawResult.(model.Vector)
-		if !ok {
-			t.Fatalf("expected prometheus query to return type Vector, got %T", resultVector)
 		}
 		metricsCollected := false
 		for _, sample := range resultVector {
@@ -687,6 +680,68 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
 	})
 	if err != nil {
 		t.Fatalf("Health check of metrics collection in namespace %s did not succeed within %v", serviceCAOperatorNamespace, timeout)
+	}
+}
+
+func runPromQueryForVector(t *testing.T, promClient prometheusv1.API, query string, sampleTime time.Time) (model.Vector, error) {
+	results, warnings, err := promClient.Query(context.Background(), query, sampleTime)
+	if err != nil {
+		return model.Vector{}, err
+	}
+	if len(warnings) > 0 {
+		t.Logf("prometheus query emitted warnings: %v", warnings)
+	}
+
+	result, ok := results.(model.Vector)
+	if !ok {
+		return model.Vector{}, fmt.Errorf("expecting vector type result, found: %v ", reflect.TypeOf(results))
+	}
+
+	return result, nil
+}
+
+func getSampleForPromQuery(t *testing.T, promClient prometheusv1.API, query string, sampleTime time.Time) (*model.Sample, error) {
+	res, err := runPromQueryForVector(t, promClient, query, sampleTime)
+	if err != nil {
+		return nil, err
+	}
+	if len(res) == 0 {
+		return nil, fmt.Errorf("no matching metrics found for query %s", query)
+	}
+	return res[0], nil
+}
+
+func checkServiceCAMetrics(t *testing.T, client *kubernetes.Clientset, promClient prometheusv1.API) {
+	timeout := 60 * time.Second
+
+	secret, err := client.CoreV1().Secrets(serviceCAControllerNamespace).Get(signingKeySecretName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("error retrieving signing key secret: %v", err)
+	}
+	currentCACerts, err := util.PemToCerts(secret.Data[v1.TLSCertKey])
+	if err != nil {
+		t.Fatalf("error unmarshaling %q: %v", v1.TLSCertKey, err)
+	}
+	if len(currentCACerts) == 0 {
+		t.Fatalf("no signing keys found")
+	}
+
+	want := currentCACerts[0].NotAfter
+	err = wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
+		rawExpiryTime, err := getSampleForPromQuery(t, promClient, `service_ca_expiry_time_seconds`, time.Now())
+		if err != nil {
+			t.Logf("failed to get sample value: %v", err)
+			return false, nil
+		}
+
+		if float64(want.Unix()) != float64(rawExpiryTime.Value) {
+			t.Fatalf("service ca expiry time mismatch expected %v observed %v", float64(want.UnixNano()), float64(rawExpiryTime.Value))
+		}
+
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("service ca expiry timer metrics collection failed: %v", err)
 	}
 }
 
@@ -877,13 +932,20 @@ func TestE2E(t *testing.T) {
 		}
 	})
 
-	// Test that the operator's metrics endpoint is being read by prometheus
 	t.Run("metrics", func(t *testing.T) {
 		promClient, err := newPrometheusClientForConfig(adminConfig)
 		if err != nil {
 			t.Fatalf("error initializing prometheus client: %v", err)
 		}
-		checkMetricsCollection(t, promClient, "openshift-service-ca-operator")
+		// Test that the operator's metrics endpoint is being read by prometheus
+		t.Run("collection", func(t *testing.T) {
+			checkMetricsCollection(t, promClient, "openshift-service-ca-operator")
+		})
+
+		// Test that service CA metrics are collected
+		t.Run("service-ca-metrics", func(t *testing.T) {
+			checkServiceCAMetrics(t, adminClient, promClient)
+		})
 	})
 
 	t.Run("refresh-CA", func(t *testing.T) {


pkg/operator/starter.go
```diff
@@ -110,6 +111,9 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		controllerContext.EventRecorder,
 	)
 
+	//register metrics
```
marun: Please omit comments that provide no info over and above the code.

test/e2e/e2e_test.go
```diff
@@ -642,6 +642,25 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
 	}
 }
 
+func checkSignerCAMetrics(t *testing.T, promClient prometheusv1.API) {
+	timeout := 60 * time.Second
+	err := wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
+		results, warnings, err := promClient.Query(context.Background(), `signer_ca_expiry_time`, time.Now())
```
marun: signer_ca -> service_ca

test/e2e/e2e_test.go
```diff
@@ -902,6 +921,16 @@ func TestE2E(t *testing.T) {
 		checkMetricsCollection(t, promClient, "openshift-service-ca-operator")
 	})
 
+	// Test that signer CA metrics are collected
+	// Test that the operator's metrics endpoint is being read by prometheus
+	t.Run("signer-ca-metrics", func(t *testing.T) {
+		promClient, err := newPrometheusClientForConfig(adminConfig)
```
marun: Given initialization of a prometheus client involves a bunch of api calls, could the 2 metrics tests share its setup?  Maybe by nesting the tests e.g.

```golang
t.Run("metrics", func(t *testing.T) {
  promClient, err := ...
  ...
  t.Run("collection", func(t *testing.T) {
    ...
  }
  t.Run("signer-ca-expiry", func(t *testing.T {
    ...
  }
}
```

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+func (s *signerCAMetrics) Describe(ch chan<- *prometheus.Desc) {
+	ch <- s.expiryTime.WithLabelValues("").Desc()
+	ch <- s.timeLeft.WithLabelValues("").Desc()
+}
+
+// Collect calculates metrics from the cached config and reports them to the prometheus collector.
+func (s *signerCAMetrics) Collect(ch chan<- prometheus.Metric) {
+	secret := resourceread.ReadSecretV1OrDie(v4_00_assets.MustAsset("v4.0.0/service-serving-cert-signer-controller/signing-secret.yaml"))
+
+	if existing, err := s.client.Secrets(secret.Namespace).Get(secret.Name, metav1.GetOptions{}); err == nil {
```
marun: How much work would it be instead to have the operator sync loop collect `NotAfter` for the current cert and communicate it the metrics instance in a safe (read: synchronized) way?

test/e2e/e2e_test.go
```diff
@@ -642,6 +642,25 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
...
+		results, warnings, err := promClient.Query(context.Background(), `signer_ca_expiry_time`, time.Now())
+		if err != nil {
+			t.Errorf("failed to execute prometheus query: %v", err)
+			return false, nil
+		}
+		if len(warnings) > 0 {
+			t.Logf("prometheus query emitted warnings: %v", warnings)
+		}
+
+		return len(results.String()) != 0, nil
```
marun: Please confirm that the values of `signer_ca_expiry_time` and `signer_ca_time_left` are as expected (i.e retrieve the signing secret in the test and check against its expiry).

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func Register(corev1Client coreclientv1.CoreV1Interface) {
+	legacyregistry.MustRegister(&signerCAMetrics{
+		client: corev1Client,
+		expiryTime: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "signer_ca_expiry_time",
+			Help: "Reports the expiry time of signer CA used in the cluster.",
```
marun: s/signer/service/

test/e2e/e2e_test.go
```diff
@@ -893,13 +955,20 @@ func TestE2E(t *testing.T) {
 		}
 	})
 
-	// Test that the operator's metrics endpoint is being read by prometheus
+	promClient, err := newPrometheusClientForConfig(adminConfig)
+	if err != nil {
+		t.Fatalf("error initializing prometheus client: %v", err)
+	}
 	t.Run("metrics", func(t *testing.T) {
-		promClient, err := newPrometheusClientForConfig(adminConfig)
```
marun: Please move the client initialization back to the test so that it will only run if the test does.

test/e2e/e2e_test.go
```diff
@@ -642,6 +642,68 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
 	}
 }
 
+func runPromQuery(t *testing.T, promClient prometheusv1.API, query string, sampleTime time.Time) (float64, error) {
```
marun: nit: Maybe use a more descriptive name? This function has the specific purpose of returning a scalar result, so maybe something like `queryForScalar`?

test/e2e/e2e_test.go
```diff
@@ -642,6 +642,68 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
...
+			return false, nil
+		}
+
+		if want.Sub(sampleTime).Seconds() != have {
+			t.Fatalf("time left for service ca expiry did not match. have %v want %v", have, want.Sub(sampleTime).Seconds())
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("service ca expiry timer metrics collection failed %v", err)
```
marun: nit: Given that the only error that will appear here is due to timeout, maybe `("Failed to verify ca expiry metrics within %v", timeout)`?

nits:

- The first word of error messages should be lower-case because error messages are frequently aggregated in log output.
- The first word of a logged message (including to test logging) should instead be capitalized because it will be output directly rather than aggregated for output.
- Please ensure separation of message from serialized error (e.g. "<message>: %v" instead of "<message> %v").

pkg/operator/starter.go
vareti: removed it now.

test/e2e/e2e_test.go
vareti: renamed instances of signer_ca to service_ca

test/e2e/e2e_test.go
vareti: now the tests run in parallel and share the setup!

test/e2e/e2e_test.go
vareti: updated unit test to check both the values are as expected.

pkg/operator/metrics/metrics.go
vareti: Done.

pkg/operator/metrics/metrics.go
vareti: I am looking into this.

test/e2e/e2e_test.go
vareti: I was trying to follow this test pattern mentioned in the golang blog where setup code is before the test https://blog.golang.org/subtests. Look for the section `Setup and Tear-down`

    func TestTeardownParallel(t *testing.T) {
        // <setup code>
        // This Run will not return until its parallel subtests complete.
        t.Run("group", func(t *testing.T) {
            t.Run("Test1", parallelTest1)
            t.Run("Test2", parallelTest2)
            t.Run("Test3", parallelTest3)
        })
        // <tear-down code>
    }

Do you still want me to move the initialization back to the test?

vareti:
I usually respond to all comments and let reviewers know once I make all the requested changes. As, I am yet to make one more change, I did not respond. Maybe from next time, I will respond as I am making changes.

test/e2e/e2e_test.go
marun: Where setup code is used by all tests it should appear outside of it. Since the setup is only used by the metrics tests, it can appear inside of the wrapping test before the subtests are defined e.g.

```golang
func TestFoo(...) {

  t.Run("metrics", func(...) {
    <setup>
    t.Run("metrics test 1", func(...) {}
    t.Run("metrics test 2", func(...) {}
  }
}
```

Note that we're not enabling parallel execution of these tests. Care would have to be taken to do so because the rotation tests aren't compatible with concurrent execution - they affect global state that other tests depend on.

test/e2e/e2e_test.go
vareti: Using these code guidelines in recent commits

test/e2e/e2e_test.go
vareti: updated. please take a look and let me know if it needs some more refactoring

test/e2e/e2e_test.go
vareti: updated the code and moved intialization inside `t.Run`

pkg/operator/metrics/metrics.go
vareti: Updated the code. Please take a look and let me if it looks better.

vareti:
@marun addressed review comments. please take a look. also e2e tests are passing locally

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	ch <- s.expiryTime.WithLabelValues("")
+	ch <- s.timeLeft.WithLabelValues("")
+}
+
+func (s *ServiceCAMetrics) ClearState() {}
+
+func (s *ServiceCAMetrics) Compute(client coreclientv1.CoreV1Interface) {
```
marun: When I asked for the current CA duration to be collected via the sync loop, I meant in `manageSignerCA` where the CA is already being parsed: 

https://github.com/openshift/service-ca-operator/blob/master/pkg/operator/sync_common.go#L77

I don't think it should be necessary for metrics computation to independently load and parse the CA.

vareti: moved logic to `manageSignerCA` and metrics package will not be aware of CA data.

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+		expiryTime: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "service_ca_expiry_time",
+			Help: "Reports the expiry time of service CA used in the cluster.",
+		}, []string{"expiry_time"}),
+		timeLeft: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "service_ca_time_left",
+			Help: "Reports the time left before the service CA expires.",
+		}, []string{"time_left"}),
+	}
+	legacyregistry.MustRegister(s)
```
sttts: constructors with side-effects are bad. Would just fallback to a simple init func here.

vareti: Thanks for pointing this out. updated the code now.

vareti:
@marun PR is updated with some changes. please take a look.

pkg/operator/sync_common.go
```diff
@@ -115,12 +116,19 @@ func manageSignerCA(client coreclientv1.SecretsGetter, eventRecorder events.Reco
 			return false, fmt.Errorf("failed to rotate signing CA: %v", err)
 		}
 		if len(rotationMsg) == 0 {
+			metrics.Compute(existingCert.NotAfter)
```
marun: I think the only place that expiry should be recorded is at the beginning of this else clause (i.e. after L113). In the event that a new certificate is generated (a rare event), the next reconcile will occur shortly after and guarantee that the expiry is recorded in a timely fashion.

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+// Collect calculates metrics from the cached config and reports them to the prometheus collector.
+func (s *ServiceCAMetrics) Collect(ch chan<- prometheus.Metric) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	ch <- s.expiryTime
+	ch <- s.timeLeft
+}
+
+func (s *ServiceCAMetrics) ClearState() {}
```
marun: Is implementing this unnecessary?

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (s *ServiceCAMetrics) Compute(notAfter time.Time) {
```
marun: nit: I think `Compute` is too vague a name. Maybe rename to something like `SetCAExpiry`?

test/e2e/e2e_test.go
```diff
@@ -690,6 +683,88 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
 	}
 }
 
+func runPromQueryForVector(t *testing.T, promClient prometheusv1.API, query string, sampleTime time.Time) (model.Vector, error) {
+	results, warnings, err := promClient.Query(context.Background(), query, sampleTime)
+	if err != nil {
+		t.Errorf("failed to execute prometheus query: %v", err)
```
marun: Why is desirable to both log and return the error?

test/e2e/e2e_test.go
```diff
@@ -690,6 +683,88 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
...
+		return model.Vector{}, err
+	}
+	if len(warnings) > 0 {
+		t.Logf("prometheus query emitted warnings: %v", warnings)
+	}
+
+	result, ok := results.(model.Vector)
+	if !ok {
+		t.Fatalf("expecting scalar type result, found: %v ", reflect.TypeOf(results))
```
marun: I think a good rule of thumb for test helpers is to either accept testing.T and use `t.Fatalf` to fail, or return an error.  Mixing the 2 error handling strategies is likely to be confusing for future maintainers.

test/e2e/e2e_test.go
```diff
@@ -690,6 +683,88 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
...
+	}
+
+	return result, nil
+}
+
+func getSampleForPromQuery(t *testing.T, promClient prometheusv1.API, query string, sampleTime time.Time) (*model.Sample, error) {
+	res, err := runPromQueryForVector(t, promClient, query, sampleTime)
+	if err != nil {
+		t.Errorf("failed to execute prometheus query: %v", err)
```
marun: nit: Given that `runPromQueryForVector` returns formatted errors, `t.Errorf(err)` is likely sufficient.  But again, is logging necessary if the error is returned?

test/e2e/e2e_test.go
```diff
@@ -690,6 +683,88 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
...
+		if err != nil {
+			t.Errorf("failed to get sample value: %v", err)
+			return false, nil
+		}
+
+		if float64(want.Unix()) != float64(have.Value) {
+			t.Fatalf("service ca expiry time mismatch expected %v observed %v", float64(want.UnixNano()), float64(have.Value))
+		}
+
+		have, err = getSampleForPromQuery(t, promClient, `service_ca_time_left`, sampleTime)
```
marun: nit: Rather than using `have` twice, please use distinct names for the destination variables (e.g. `rawExpiryTime`, `rawTimeLeft`).

test/e2e/e2e_test.go
```diff
@@ -690,6 +683,88 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
...
+		if err != nil {
+			t.Errorf("failed to get sample value: %v", err)
+			return false, nil
+		}
+		// I could not find a direct way to check time left before expiry
```
marun: Given how often metrics are collected, the reported and computed time should be within, say, 5 minutes of one another (e.g. `((expiryTime - now) - timeLeft) < 5 * duration.Minute`).  Given that the expiry duration of a fresh test cluster will be 14 months, being super accurate isn't critical.

marun:
@openshift/openshift-team-monitoring These metrics would ideally be collected via telemeter for all clusters so that we have visibility into the effectiveness of automated service ca rotation.

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/blang/semver"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"sync"
+	"time"
+)
+
+// ServiceCAMetrics implements metrics gathering for this component.
+type ServiceCAMetrics struct {
+	lock       sync.Mutex
```
brancz: There should be no need to lock any of this. The metrics are set atomically and you should just register the metrics directly into the metrics registry.

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func NewServiceCAMetrics() *ServiceCAMetrics {
+	return &ServiceCAMetrics{
+		expiryTime: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "service_ca_expiry_time",
+			Help: "Reports the expiry time of service CA used in the cluster.",
+		}),
+		timeLeft: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "service_ca_time_left",
```
simonpasquier: this metric isn't really needed since it can be computed by Prometheus directly: `service_ca_expiry_time - time()`

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+type ServiceCAMetrics struct {
+	lock       sync.Mutex
+	expiryTime prometheus.Gauge
+	timeLeft   prometheus.Gauge
+}
+
+func NewServiceCAMetrics() *ServiceCAMetrics {
+	return &ServiceCAMetrics{
+		expiryTime: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "service_ca_expiry_time",
```
simonpasquier: `service_ca_expiry_time_seconds` would be better

pkg/operator/metrics/metrics.go
marun: Metrics in this instance are being written by one goroutine (the operator worker) and read by another (the server exposing the metrics for collection). That suggests synchronization to me. Is there another strategy for ensuring safe concurrent access to the state variables that you would suggest?

brancz: Yes just create and register the metrics once, the registration process is already concurrency safe.

marun: I'm not following. Can you point me to code that is ensuring that concurrent access to the state variables will be synchronized?

brancz: Look at this example app for example https://github.com/brancz/prometheus-example-app/blob/master/main.go

vareti: @brancz Thanks for the review. I understand that mutex is not required if we are only using one metric because of the atomicity. But if there are multiple metrics that are inter-dependent, don't we need a mutex in that case?

marun: Gotcha. As you were trying to say, `Gauge.Set` ensures synchronization by setting the value atomically.

pkg/operator/metrics/metrics.go
vareti: :+1:

pkg/operator/metrics/metrics.go
vareti: Thanks for the review. I removed the extra metric

pkg/operator/sync_common.go
vareti: As I understand it, at the beginning of else block, it is still not certain if the certificate would be re-generated. That will be certain only after `maybeRotateSigningSecret` is executed. That is one case.

There is another case which is handled in the if block where a certificate is not found in the cluster. A new certificate would be generated in this case too. 

So, I think the best place to compute the metric is after the if/else statement. Sounds good?

vareti: That could be the only place where expiry time can be computed if there is no need to update expiry time for every reconcile loop, which actually makes sense.

pkg/operator/metrics/metrics.go
vareti: No. This is need by the `Registerable` interface used when metric is registered using `legacyregistry.MustRegister`

vareti: https://github.com/kubernetes/component-base/blob/v0.17.0/metrics/registry.go#L106

pkg/operator/metrics/metrics.go
vareti: Yes. Better!

test/e2e/e2e_test.go
vareti: Yes, you are right! maybe because of my habit of over-logging things. Modified it to return the error now instead of printing it.

test/e2e/e2e_test.go
vareti: Handled in the way mentioned in my previous comment

test/e2e/e2e_test.go
vareti: Same! handled as i said in my previous comments.

test/e2e/e2e_test.go
vareti: updated the code

test/e2e/e2e_test.go
vareti: We don't have this metric anymore.

pkg/operator/metrics/metrics.go
vareti: Okay! My page isn't refreshed and I didn't see your conversation.

vareti: But my question is still open?

vareti:
@marun thanks for the review. addressed your comments. please take a look.

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+	return &ServiceCAMetrics{
+		expiryTime: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "service_ca_expiry_time_seconds",
+			Help: "Reports the expiry time of service CA used in the cluster.",
+		}),
+	}
+}
+
+func (s *ServiceCAMetrics) Register() {
+	legacyregistry.MustRegister(s)
```
brancz: Just register the metric directly then ServiceCAMetrics doesn’t need to implement the collector interface unnecessarily.

vareti: @brancz Thanks for the tip. I already tried registering the metric directly `prometheus.MustRegister` but it did not work. I could not find the metric in web console when I queried for it. It is possible that I might be doing something wrong and I did not debug it further. I will take a look again.

brancz:
Lgtm now

test/e2e/e2e_test.go
```diff
@@ -690,6 +683,68 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
 	}
 }
 
+func runPromQueryForVector(t *testing.T, promClient prometheusv1.API, query string, sampleTime time.Time) (model.Vector, error) {
+	results, warnings, err := promClient.Query(context.Background(), query, sampleTime)
+	if err != nil {
+		return model.Vector{}, err
```
marun: nit: Please return nil. Best practice is to return the 'zero' value of return values when returning an error, and that value for `Vector` is `nil` due to it being an alias for a slice type.

test/e2e/e2e_test.go
```diff
@@ -690,6 +683,68 @@ func checkMetricsCollection(t *testing.T, promClient prometheusv1.API, namespace
...
+	}
+	if len(warnings) > 0 {
+		t.Logf("prometheus query emitted warnings: %v", warnings)
+	}
+
+	result, ok := results.(model.Vector)
+	if !ok {
+		return model.Vector{}, fmt.Errorf("expecting vector type result, found: %v ", reflect.TypeOf(results))
```
marun: nit: return nil

vareti:
/hold

There seems to be an issue with the new e2e test. Let me look into it.

pkg/operator/metrics/metrics.go
```diff
@@ -0,0 +1,21 @@
...
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+var expiryTime = metrics.NewGauge(&metrics.GaugeOpts{
+	Name: "service_ca_expiry_time_seconds",
+	Help: "Reports the expiry time of service CA used in the cluster.",
+})
+
+func init() {
+	legacyregistry.MustRegister(expiryTime)
```
s-urbaniak: just a nit: instead of having a global registry, maybe think about dependency injecting the prometheus registry.

PR #97, opened 2020-01-21
Remove unnecessary permissions for controllers

marun:
The [controller unification PR](https://github.com/openshift/service-ca-operator/pull/89) added permissions for removing 4.3 resources to the role used by controllers but removal will be performed by the operator which has cluster admin privileges.

/cc @stlaz


marun:
Rebased

stlaz:
/lgtm
I should have noticed that when I pointed out the perms weren't enough to remove the deployment

PR #98, opened 2020-01-21
e2e test to check service serving cert using client pod

vareti:
Adds an e2e test to check service serving certificate is recreated if the secret is deleted and it matches the certificate received by the client pod

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -52,7 +52,7 @@ const (
 	rotationTimeout = 3 * time.Minute
 	// Polling for resources related to rotation may be delayed by the number of resources
 	// that are updated in the cluster in response to rotation.
-	rotationPollTimeout = 60 * time.Second
+	rotationPollTimeout = 2 * time.Minute
 )
 
 // checkComponents verifies that the components of the operator are running.
@@ -589,6 +589,100 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
 	return obj, err
 }
 
+func checkClientPodRcvdUpdatedServerCert(t *testing.T, client *kubernetes.Clientset, service *v1.Service, testNS, updatedServerCert string) {
+	timeout := 5 * time.Minute
+	metricsURL := fmt.Sprintf("%s.%s.svc:%d", service.Name, service.Namespace, service.Spec.Ports[0].Port)
+	err := wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
+		podName := "client-pod-" + randSeq(5)
+		_, err := client.CoreV1().Pods(testNS).Create(&v1.Pod{
+			TypeMeta: metav1.TypeMeta{},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      podName,
+				Namespace: testNS,
+			},
+			Spec: v1.PodSpec{
+				Containers: []v1.Container{
+					{
+						Name:    "cert-checker",
+						Image:   "nicolaka/netshoot:latest",
+						Command: []string{"/bin/bash"},
+						Args:    []string{"-c", fmt.Sprintf("openssl s_client -showcerts -connect %s < /dev/null 2>/dev/null | openssl x509", metricsURL)},
+					},
+				},
+				RestartPolicy: v1.RestartPolicyOnFailure,
+			},
+		})
+		if err != nil {
+			t.Logf("creating client pod failed: %v", err)
+			return false, nil
+		}
+		defer deletePod(t, client, podName, testNS)
+
+		err = waitForPodPhase(t, client, podName, testNS, v1.PodSucceeded)
+		if err != nil {
+			t.Logf("wait on pod to complete failed: %v", err)
+			return false, nil
+		}
+
+		serverCertClientReceived, err := getPodLogs(t, client, podName, testNS)
+		if err != nil {
+			t.Logf("fetching pod logs failed: %v", err)
+			return false, nil
+		}
+		return strings.Contains(updatedServerCert, serverCertClientReceived), nil
+	})
+	if err != nil {
+		t.Fatalf("failed to verify updated certs within timeout(%v)", timeout)
+	}
+
+}
+
+func waitForPodPhase(t *testing.T, client *kubernetes.Clientset, name, namespace string, phase v1.PodPhase) error {
+	return wait.PollImmediate(10*time.Second, time.Minute, func() (bool, error) {
+		pod, err := client.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			t.Logf("fetching test pod from apiserver failed: %v", err)
+			return false, nil
+		}
+		return pod.Status.Phase == phase, nil
+	})
+}
+
+func getPodLogs(t *testing.T, client *kubernetes.Clientset, name, namespace string) (string, error) {
+	rc, err := client.CoreV1().Pods(namespace).GetLogs(name, &v1.PodLogOptions{}).Stream()
+	if err != nil {
+		return "", err
+	}
+	defer rc.Close()
+
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(rc)
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+func deletePod(t *testing.T, client *kubernetes.Clientset, name, namespace string) {
+	err := client.CoreV1().Pods(namespace).Delete(name, &metav1.DeleteOptions{})
+	if errors.IsNotFound(err) {
+		return
+	}
+	if err != nil {
+		t.Errorf("failed to delete pod: %v", err)
+	}
+}
+
+func deleteNamespace(t *testing.T, client *kubernetes.Clientset, name string) {
+	err := client.CoreV1().Namespaces().Delete(name, &metav1.DeleteOptions{})
+	if errors.IsNotFound(err) {
+		return
+	}
+	if err != nil {
+		t.Errorf("failed to delete namespace: %v", err)
+	}
+}
+
 // newPrometheusClientForConfig returns a new prometheus client for
 // the provided kubeconfig.
 func newPrometheusClientForConfig(config *rest.Config) (prometheusv1.API, error) {
@@ -765,6 +859,36 @@ func TestE2E(t *testing.T) {
 		if !bytes.Equal(originalBytes, updatedBytes) {
 			t.Fatalf("did not expect TLSCertKey to be replaced with a new cert")
 		}
+	})
+
+	// make sure that deleting service-cert-secret regenerates a working secret again
+	t.Run("serving-cert-secret-delete-data", func(t *testing.T) {
+		serviceName := "metrics"
+		operatorNamespace := "openshift-service-ca-operator"
+		testNamespace := "test-" + randSeq(5)
+		_, err := createTestNamespace(adminClient, testNamespace)
+		if err != nil {
+			t.Fatalf("could not create test namespace: %v", err)
+		}
+		defer deleteNamespace(t, adminClient, testNamespace)
+		service, err := adminClient.CoreV1().Services(operatorNamespace).Get(serviceName, metav1.GetOptions{})
+		if err != nil {
+			t.Fatalf("fetching service from apiserver failed: %v", err)
+		}
+		secretName, ok := service.ObjectMeta.Annotations[api.ServingCertSecretAnnotation]
+		if !ok {
+			t.Fatalf("secret name not found in service annotations")
+		}
+		err = adminClient.CoreV1().Secrets(operatorNamespace).Delete(secretName, &metav1.DeleteOptions{})
+		if err != nil {
+			t.Fatalf("deleting secret %s in namespace %s failed: %v", secretName, operatorNamespace, err)
+		}
+		updatedBytes, _, err := pollForUpdatedServingCert(t, adminClient, operatorNamespace, secretName, rotationPollTimeout, nil, nil)
+		if err != nil {
+			t.Fatalf("error fetching re-created serving cert secret: %v", err)
+		}
+
+		checkClientPodRcvdUpdatedServerCert(t, adminClient, service, testNamespace, string(updatedBytes))
 	})
 
 	// test ca bundle injection configmap


test/e2e/e2e_test.go
```diff
@@ -93,8 +98,35 @@ func createTestNamespace(client *kubernetes.Clientset, namespaceName string) (*v
 	return ns, err
 }
 
+func createServiceEndpoint(client *kubernetes.Clientset, podName, namespace string, labels map[string]string) error {
```
marun: Please use the metrics endpoint of the service ca operator rather than creating a new pod.

test/e2e/e2e_test.go
```diff
@@ -591,6 +624,70 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
 	return obj, err
 }
 
+func pollForUpdatedServerCertFromClient(t *testing.T, client *kubernetes.Clientset, config *rest.Config, namespace, serviceName, updatedServerCert string) error {
```
marun: Rather than creating a pod and reading data from it, how about the following?

- use alpine or busybox
- construct a command for the pod that directly tests for the expected cert
  - e.g. `bash -c "test <expected cert> == $(command to retrieve cert)"`
- check the exit code of the pod

Simpler and a good opportunity to exercise your bash-foo. :)

test/e2e/e2e_test.go
```diff
@@ -769,6 +866,50 @@ func TestE2E(t *testing.T) {
...
+		ns, err := createTestNamespace(adminClient, "test-"+randSeq(5))
+		if err != nil {
+			t.Fatalf("could not create test namespace: %v", err)
+		}
+		testServiceName := "test-service-" + randSeq(5)
+		testSecretName := "test-secret-" + randSeq(5)
+		labels := map[string]string{
+			"app": "test",
+		}
+		defer cleanupServiceSignerTestObjects(adminClient, testSecretName, testServiceName, ns.Name)
```
marun: Usually cleanup functions are called after successful creation rather than before.

test/e2e/e2e_test.go
vareti: pushed changes to use metrics endpoint. please take a look.

test/e2e/e2e_test.go
vareti: I copied it from the test above. anyways it need not be addresses as it uses metrics endpoint of service-ca-operator now.

test/e2e/e2e_test.go
vareti: > use alpine or busybox

maybe i did not search enough and this could done with alpine or busybox images. from my understanding, `openssl` is a good way to get server certs. I am not sure how to do this with other tools. Any pointers are very welcome 

> construct a command for the pod that directly tests for the expected cert

I am looking into this

vareti:
@marun Thanks for the review. Gave my comments and posted some questions. Please take a look

test/e2e/e2e_test.go
vareti: I pushed some changes. it does not entirely depend on bash script but will use a blend of both of bash commands and pod logs. please review

vareti:
@marun it is ready for review one more time. please take a look

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
+			Name:      podName,
+			Namespace: service.Namespace,
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:    "cert-checker",
+					Image:   "nicolaka/netshoot:latest",
+					Command: []string{"/bin/bash"},
+					Args:    []string{"-c", fmt.Sprintf("true | openssl s_client -showcerts -connect %s 2>/dev/null | openssl x509", metricsURL)},
```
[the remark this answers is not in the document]
vareti: `openssl s_client` waits to send any user given characters through stdin to remote server. Using `true |` we are signalling termination of command.

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
 
+func pollForUpdatedServerCertFromClient(t *testing.T, client *kubernetes.Clientset, service *v1.Service, updatedServerCert string) error {
+	podName := "client-pod"
+	metricsURL := fmt.Sprintf("%s.%s.svc:%d", service.Name, service.Namespace, service.Spec.Ports[0].Port)
+	defer client.CoreV1().Pods(service.Namespace).Delete(podName, &metav1.DeleteOptions{})
+	_, err := client.CoreV1().Pods(service.Namespace).Create(&v1.Pod{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: service.Namespace,
```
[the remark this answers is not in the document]
vareti: done. moved them to test namespace.

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
 	return obj, err
 }
 
+func pollForUpdatedServerCertFromClient(t *testing.T, client *kubernetes.Clientset, service *v1.Service, updatedServerCert string) error {
+	podName := "client-pod"
+	metricsURL := fmt.Sprintf("%s.%s.svc:%d", service.Name, service.Namespace, service.Spec.Ports[0].Port)
+	defer client.CoreV1().Pods(service.Namespace).Delete(podName, &metav1.DeleteOptions{})
```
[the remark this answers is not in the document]
vareti: missed them! now there are functions to delete pod and namespace during clean up

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
+					Command: []string{"/bin/bash"},
+					Args:    []string{"-c", fmt.Sprintf("true | openssl s_client -showcerts -connect %s 2>/dev/null | openssl x509", metricsURL)},
+				},
+			},
+		},
+	})
+	if err != nil {
+		t.Fatalf("creating client pod failed: %v", err)
+	}
+	timeout := 5 * time.Minute
```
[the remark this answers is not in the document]
vareti: worst case is around 5 minutes. typically operation completes in a minute.

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
+	err = wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
+		req := client.CoreV1().Pods(service.Namespace).GetLogs(podName, &v1.PodLogOptions{})
+		podLogs, err := req.Stream()
+		if err != nil {
+			t.Logf("Could not get pod logs: %v", err)
```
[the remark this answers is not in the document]
vareti: This is a good point I did not consider. Now I updated the code to look for logs after pod moves to complete state.

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
+		}
+		defer podLogs.Close()
+
+		buf := new(bytes.Buffer)
+		_, err = io.Copy(buf, podLogs)
+		if err != nil {
+			t.Logf("Error in copying bytes to a buffer: %v", err)
+		}
+		serverCertClientReceived := buf.String()
+		if strings.Contains(updatedServerCert, serverCertClientReceived) {
```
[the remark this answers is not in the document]
vareti: `updateServerCert` contains the certificate chain. `serverCertClientReceived` contains the leaf certificate that client looks at. Both may not be equal.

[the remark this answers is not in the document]
vareti: Updated the return statement as well.

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
 	return obj, err
 }
 
+func pollForUpdatedServerCertFromClient(t *testing.T, client *kubernetes.Clientset, service *v1.Service, updatedServerCert string) error {
```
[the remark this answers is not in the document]
vareti: you are right! it does not make sense if we are not checking for error. I updated the code now to not return any error now.

test/e2e/e2e_test.go
```diff
@@ -769,6 +821,36 @@ func TestE2E(t *testing.T) {
...
+		}
+		secretName, ok := service.ObjectMeta.Annotations[api.ServingCertSecretAnnotation]
+		if !ok {
+			t.Fatalf("secret name not found in service annotations")
+		}
+		err = adminClient.CoreV1().Secrets(namespace).Delete(secretName, &metav1.DeleteOptions{})
+		if err != nil {
+			t.Fatalf("deleting secret %s in namespace %s failed: %v", secretName, namespace, err)
+		}
+		err = pollForServiceServingSecret(adminClient, secretName, namespace)
```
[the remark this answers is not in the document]
vareti: yes. used the suggested and it simplified the code by one function call.

test/e2e/e2e_test.go
```diff
@@ -591,6 +592,57 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: service.Namespace,
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:    "cert-checker",
+					Image:   "nicolaka/netshoot:latest",
```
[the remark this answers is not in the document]
stevekuznetsov: We don't have any policy that would disallow it. I'd suggest maybe adding an ImageStream to `openshift/release/core-services/addiitional-images/` so that we don't need to pull it

test/e2e/e2e_test.go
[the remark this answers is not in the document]
Miciah: A more idiomatic solution is to redirect stdin to /dev/null: `openssl s_client ... </dev/null`.

test/e2e/e2e_test.go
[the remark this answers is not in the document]
vareti: @stevekuznetsov thanks for the reply. I don't see `openshift/release/core-services/addiitional-images/` directory in release repo. Am i missing something?

test/e2e/e2e_test.go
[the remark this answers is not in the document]
vareti: Thanks @Miciah for the suggestion. I will update the code.

vareti:
@marun did some clean up and pushed some changes. Could you please review

test/e2e/e2e_test.go
```diff
@@ -589,6 +589,96 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
+						Name:    "cert-checker",
+						Image:   "nicolaka/netshoot:latest",
+						Command: []string{"/bin/bash"},
+						Args:    []string{"-c", fmt.Sprintf("openssl s_client -showcerts -connect %s < /dev/null 2>/dev/null | openssl x509", metricsURL)},
+					},
+				},
+				RestartPolicy: v1.RestartPolicyOnFailure,
+			},
+		})
+		defer deletePod(client, podName, testNS)
```
marun: (No action required) Not necessary to schedule deletion until after the check for an error.

test/e2e/e2e_test.go
```diff
@@ -589,6 +589,96 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
 	return obj, err
 }
 
+func checkClientPodRcvdUpdatedServerCert(t *testing.T, client *kubernetes.Clientset, service *v1.Service, testNS, updatedServerCert string) {
+	timeout := 5 * time.Minute
+	metricsURL := fmt.Sprintf("%s.%s.svc:%d", service.Name, service.Namespace, service.Spec.Ports[0].Port)
+	err := wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
+		podName := "client-pod-" + randSeq(5)
+		_, err := client.CoreV1().Pods(testNS).Create(&v1.Pod{
```
marun: Why is it necessary to create the pod in the wait loop rather than outside it?

test/e2e/e2e_test.go
```diff
@@ -589,6 +589,96 @@ func pollForResource(t *testing.T, resourceID string, timeout time.Duration, acc
...
+		return strings.Contains(updatedServerCert, serverCertClientReceived), nil
+	})
+	if err != nil {
+		t.Fatalf("failed to verify updated certs within timeout(%v)", timeout)
+	}
+
+}
+
+func waitForPodPhase(t *testing.T, client *kubernetes.Clientset, name, namespace string, phase v1.PodPhase) error {
+	return wait.PollImmediate(10*time.Second, time.Minute, func() (bool, error) {
```
marun: Nesting wait loops is something of an anti-pattern. Is it necessary in this case?

test/e2e/e2e_test.go
marun: As per a previous comment, it's still necessary to check error conditions when calling something with `defer`. As written here, any error will be ignored. At the least, it should be logged:

```golang
defer func() {
  err := deletePod(...)
  if err != nil {
  ...
  }
}
```

Another option - have the deferred method accept `*testing.T` and log/fail/whatever directly.

test/e2e/e2e_test.go
```diff
@@ -767,6 +857,39 @@ func TestE2E(t *testing.T) {
 		}
 	})
 
+	// make sure that deleting service-cert-secret regenerates a working secret again
+	t.Run("serving-cert-secret-delete-data", func(t *testing.T) {
+		serviceName := "metrics"
+		namespace := "openshift-service-ca-operator"
```
marun: Given that there are 2 namespace vars in this closure, please differentiate them (e.g. `operatorNamespace`, `testNamespace`).

test/e2e/e2e_test.go
```diff
@@ -767,6 +857,39 @@ func TestE2E(t *testing.T) {
...
+		ns, err := createTestNamespace(adminClient, "test-"+randSeq(5))
```
marun: Given that only the name of the created namespace is used, maybe the following?

```golang
testNamespaceObj, err := ...
testNamespace := testNamespaceObj.Name
```

test/e2e/e2e_test.go
```diff
@@ -767,6 +857,39 @@ func TestE2E(t *testing.T) {
...
+		}
+		oldCertPEM, oldKeyPEM, err := pollForUpdatedServingCert(t, adminClient, namespace, secretName, rotationPollTimeout, nil, nil)
+		if err != nil {
+			t.Fatalf("error retrieving service cert: %v", err)
+		}
+		err = adminClient.CoreV1().Secrets(namespace).Delete(secretName, &metav1.DeleteOptions{})
+		if err != nil {
+			t.Fatalf("deleting secret %s in namespace %s failed: %v", secretName, namespace, err)
+		}
+		updatedBytes, _, err := pollForUpdatedServingCert(t, adminClient, namespace, secretName, rotationPollTimeout, oldCertPEM, oldKeyPEM)
```
marun: Since the secret has been deleted, it's enough to supply `nil` for `oldCertValue` and `oldKeyValue` (obviating the need for the previous call to `pollForUpdatedServingCert` on line 877.

test/e2e/e2e_test.go
vareti: Done!

test/e2e/e2e_test.go
vareti: it might often be the case that service might be serving stale certificates. so, creating pods in a loop will make sure that it retries until the new certificate is received.

test/e2e/e2e_test.go
vareti: this is something that out of our previous discussion of waiting till the pod is ready before the test fetches the logs. it simplifies the read for pod logs.

test/e2e/e2e_test.go
vareti: This is my stupidity! corrected it now!

test/e2e/e2e_test.go
vareti: Done

test/e2e/e2e_test.go
vareti: Done!

test/e2e/e2e_test.go
vareti: i was confused and added this. i understand that it is not needed and removed it.

marun:
Once this merges, the next step will be refactoring for inclusion in library-go so other operators can add tests of their rotation compatibility.

test/e2e/e2e_test.go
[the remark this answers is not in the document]
stevekuznetsov: I typo'd that, sorry -- it's at `core-services/supplemental-ci-images/` in the `release` repo.

marun:
I just remembered I have permission to merge to this repo!

/hold cancel

PR #100, opened 2020-01-23
Enable injection of service CA bundle for CRDs

marun:


marun:
@stlaz @sttts I've split the CRD injection from the webhook injection because I don't think there's anything controversial about the CRD case.

PR #102, opened 2020-01-24
Update new bundle injection targets to only support beta annotation

marun:
Only the beta ca bundle injection annotation (`service.beta.openshift.io/inject-cabundle`) should be supported for types added in 4.4. Existing types will continue to support both the alpha and beta annotations.

### DIFF
--- a/pkg/controller/cabundleinjector/starter.go
+++ b/pkg/controller/cabundleinjector/starter.go
@@ -7,12 +7,12 @@ import (
 
 	"monis.app/go/openshift/controller"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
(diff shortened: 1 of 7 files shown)


marun:
Implemented in response to https://github.com/openshift/enhancements/pull/184/files#r370656424

/cc @stlaz @sttts

pkg/controller/cabundleinjector/starter.go
```diff
@@ -91,3 +96,13 @@ func StartCABundleInjector(ctx context.Context, controllerContext *controllercmd
 
 	return nil
 }
+
+func hasSupportedInjectionAnnotation(obj v1.Object, supportedAnnotations []string) bool {
+	annotations := obj.GetAnnotations()
+	for _, key := range supportedAnnotations {
+		if annotations[key] == "true" {
```
Miciah: `HasInjectCABundleAnnotation` used `strings.EqualFold`; is it all right to replace that with `==`?

marun: Good catch! Fixed by #103

PR #109, opened 2020-02-21
Bug 1806915: remove runlevel=1 from service-ca-operator owned namespaces

stlaz:


stlaz:
/retry
interesting, SCC admission fail, aside from other operators failing as well

stlaz:
/hold
testing with the operand's SA being able to access all SCCs

bindata/v4.0.0/controller/role.yaml
```diff
@@ -4,6 +4,14 @@ metadata:
...
   namespace: openshift-service-ca
 rules:
+- apiGroups:
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
+  resourceNames:
+  - restricted
+  verbs:
+  - use
```
sttts: "use" is a verb?

stlaz: yes, it's used in the SCC admission: https://github.com/openshift/origin/blob/3c26fadaf29adfeefab38907792ebb6167526d3c/vendor/github.com/openshift/apiserver-library-go/pkg/securitycontextconstraints/sccmatching/matcher.go#L74-L74

stlaz:
/retest

The dependency loop eventually got down to cluster-policy-controller not being able to connect to KAS.

https://github.com/openshift/cluster-etcd-operator/pull/233 removes the dependency on DNS, let's see if it can be as good as solving our little dependency loop problem.

stlaz:
this won't solve the problem with upgrades though, that needs library-go fixing, because `ApplyNamespace` only merges NS meta instad of actually applying it.

stlaz:
/remove-lifecycle stale
in theory, this could now work

stlaz:
/retest
infra? AWS internal errors?

stlaz:
/test e2e-aws-operator
yay, the control plane came up

PR #110, opened 2020-03-05
Bug 1810036: Ensure service CA certs are created with unique serial numbers

marun:
This change ensures that service CA certs are generated with a unique serial number so that serving cert bundles can be loaded without error. The library-go bump ensures that the signing CA will have a randomly-generated serial number and the operator change ensures that intermediate CAs will the same.

Previously, intermediate CA certs were created with the same serial number as their template. Since a serving cert bundle includes the issuing CA cert and an intermediate CA cert (created by signing the issuing CA cert with the previous CA's private key), this lack of serial number differentiation resulted in SEC_ERROR_REUSED_ISSUER_AND_SERIAL when the bundle was read by curl due to the issuing and intermediate CAs sharing the same issuer and serial number.


pkg/operator/rotate.go
```diff
@@ -192,6 +193,19 @@ func createIntermediateCACert(targetCACert, signingCACert *x509.Certificate, sig
 	// Enable key identity chaining
 	template.AuthorityKeyId = signingCACert.SubjectKeyId
 
+	// Set a new serial number so that the intermediate CA cert is
+	// differentiated from the target CA cert. This ensures that a serving
+	// cert bundle that includes the issuing CA cert and an intermediate CA
+	// cert generated by this function - with the issuing CA cert as the
+	// target and the previous CA as the signer - will not result in
+	// SEC_ERROR_REUSED_ISSUER_AND_SERIAL when read by applications like curl.
```
stlaz: for the record: this is not about curl, curl itself would happily work if it was compiled with a different crypto backend than NSS (the crypto backend that Mozilla develops and uses in its software). https://tools.ietf.org/html/rfc5280#section-4.1.2.2 requires the serial number to be unique for each certificate signed by the CA (which this intermediate CA cert is), and NSS crypto works better when performing these checks.

pkg/operator/rotate.go
```diff
@@ -192,6 +193,19 @@ func createIntermediateCACert(targetCACert, signingCACert *x509.Certificate, sig
...
+	serialGenerator := crypto.RandomSerialGenerator{}
+	serial, err := serialGenerator.Next(template)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find next serial number: %v", err)
+	}
+	template.SerialNumber = big.NewInt(serial)
```
stlaz: We'll need the same for the signing CA as well

marun:
/hold

Need to ensure self-signed CA also has a unique serial.

pkg/operator/rotate.go
marun: I don't disagree, I was just pointing out the symptom that is being addressed by the change. The adherence to any given RFC varies according to implementation - golang doesn't care, but NSS does.

marun:
@stlaz Updated with library-go fix, PTAL.

pkg/operator/rotate.go
stlaz: Sure, my only point was - curl is not a good example as it's usually compiled with OpenSSL backend in most distributions that I've seen, so you wouldn't be able to observe the symptoms.

PR #111, opened 2020-03-09
[release-4.4] Bug 1810418: Ensure service CA certs are created with unique serial numbers  

marun:
Cherry-pick of https://github.com/openshift/service-ca-operator/pull/110

It was also necessary to include the commit that transitions to build-machinery-go since that landed in library-go's release-4.4 branch.


marun:
Updated to include real bump, PTAL

PR #115, opened 2020-04-23
Bug 1826230: kubernetes 1.18 rebase

vareti:
Rebase the operator against kubernetes 1.18, update the vendor directories and fixed any compile errors due to rebase


vareti:
operator e2e failures seems to be genuine. I will look into them.

/test e2e-aws

pkg/operator/sync_common.go
```diff
@@ -204,7 +206,12 @@ func manageDeployment(client appsclientv1.AppsV1Interface, eventRecorder events.
 	required := resourceread.ReadDeploymentV1OrDie(v4_00_assets.MustAsset(resourcePath + "deployment.yaml"))
 	required.Spec.Template.Spec.Containers[0].Image = os.Getenv("CONTROLLER_IMAGE")
 	required.Spec.Template.Spec.Containers[0].Args = append(required.Spec.Template.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", loglevel.LogLevelToVerbosity(options.Spec.LogLevel)))
-	deployment, mod, err := resourceapply.ApplyDeployment(client, eventRecorder, required, resourcemerge.ExpectedDeploymentGeneration(required, options.Status.Generations), forceDeployment)
+
+	if forceDeployment {
```
marun, line 210: Please separate this non-context change into its own commit. A general rule when bumping is to commit similar changes together to make it easier to figure out why the change was made.

Also, forcing deployment in this way is not really the intent of the change to ApplyDeployment. Better to use `ApplyDeploymentWithForce` for now: 

https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L78

The service ca sync loop is less than ideal in its current incarnation - ideally it would add the hash of the collected resourceVersions of resources it depends on as an annotation rather than deciding itself whether the rollout needed to be forced - but that can wait for a separate PR.

marun:
@p0lyn0mial voila!

p0lyn0mial:
thanks :)

marun:
Thank you!

/retitle Bug 1826230 kubernetes 1.18 rebase
/approve
/lgtm

PR #121, opened 2020-08-27
move controller boilerplate to lib-go controller factory

stlaz:
Let's see how bad this explodes.

Some of the new boilerplate with this change looks like it would find a better spot in the lib-go code instead.


marun (commented):
@standa Nice work! And since its already passing CI, I see nothing major that would prevent merge for 4.6 save a bz. Maybe it's worth filing one? Being on library-go instead of mo's library would seem like a win from a supportability perspective.

For future, I encourage you to carve up big PRs like this into functional pieces that can be submitted and reviewed independently. I get that the PR evolved over time from your original intent, but I don't think that precludes evolving your proposal strategy along with it.

/lgtm
/approve

marun:
@stlaz btw, I didn't review the unit test changes, it's just too complicated. I'd rather see them removed given that we have good e2e coverage and it's so complicated to maintain unit tests of controllers like these to limited value - most of the testing ends up being of the library-go functions.

PR #138, opened 2021-01-13
Update image tag for deployment manifest

kylape:
I am not sure if this gets automatically updated somehow, but making this change allowed me to manually deploy the operator.


marun:
@kylape The image is templated by the CVO, which will set the image according to the release payload. I don't think there is a good reason to merge this change for your local use case.

PR #142, opened 2021-02-20
Fix (make verify) with Go 1.16

mtrmac:
failing with
```
go vet -mod=vendor ./...
# github.com/openshift/service-ca-operator/test/util
test/util/rotate.go:113:4: call to (*T).Fatalf from a non-test goroutine
```

Aborting the test from inside a goroutine is invalid; instead, just mark the test as failed, and let the client on the main goroutine time out. By default, that waits too long and the test fails with a generic "timed out" panic, and no record of the cause, so impose a 1-minute connection/client timeout to ensure the test fails in time; 1 minute should be plenty.

We could do more work to fail immediately (e.g. by moving the client into a separate goroutine and having the main test goroutine wait for either of the client and server), but this seems good enough for an error we don't expect to happen.

### DIFF
--- a/test/util/rotate.go
+++ b/test/util/rotate.go
@@ -14,7 +14,7 @@ import (
 	"testing"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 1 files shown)


stlaz:
/lgtm
Thanks for the fix :+1:

PR #143, opened 2021-02-20
Don't ignore the minTimeLeft parameter to requiresRegeneration

mtrmac:
This should not change behavior, because the only non-test caller is passing in the same value.

Then fix tests that looked like they were using 1-hour time window, but actually depended on a default-initialized zero value.


marun (approved):
I appreciate your willingness to try to clean up these tests, but I would not characterize them (the existing tests) as particularly valuable (and definitely not a pattern worth repeating). A rewrite would make more sense but I would save that for a major change in functionality because of how invasive it would likely be to refactor things properly.

mtrmac:
This was created as a by-product of the new feature PR, I just wanted to get this reviewed/merged separately.

Issue #25, opened 2019-03-04
Statefulset headless services should contain a wildcard ALT-NAME


rgordill:
According with https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-identity, an statefulset can use a headless service to identify each pod. To cover this use case when generating a cert for a headless service, other than the current subject alt-names, it will be useful to have a wildcard for the pod names. Example:

**DNS Name: &ast;.broker-amq-headless.broker.svc**
DNS Name: broker-amq-headless.broker.svc 
DNS Name: broker-amq-headless.broker.svc.cluster.local

mrogers950:
@openshift/sig-auth Thoughts on this RFE? If we wanted to support it I would suggest a separate service annotation i.e.,  `service.beta.openshift.io/service-serving-wildcard`

enj:
> @openshift/sig-auth Thoughts on this RFE? If we wanted to support it I would suggest a separate service annotation i.e., `service.beta.openshift.io/service-serving-wildcard`

Not in 4.0.  Also not sure if wildcard is safe (do not have enough info to say one way or another).

rgordill:
What other alternative may cover the use case?

aaronmcohen:
Please consider this enhancement for the next release of OCP.

aaronmcohen:
Please consider this enhancement for the next release of OCP.

aaronmcohen:
Please consider this enhancement for the next release of OCP.

mtrmac:
RFC: https://github.com/openshift/enhancements/pull/674 .

A wildcard seems too broad; it makes it impossible to guarantee individual pod identities within a StatefulSet (or even to differentiate between different StatefulSets attached to a single service), and https://github.com/kubernetes/dns/blob/master/docs/specification.md only specifies what DNS records must exist, not what must not exist, so such a certificate could match unwanted DNS entries.

mtrmac:
This was implemented in #149, for headless services only.

PR #144, opened 2021-02-23
WIP: Generate per-pod certificates for StatefulSets

mtrmac:
(WIP: This seems to work well enough in manual testing, but automated tests are not implemented yet; filing now for early feedback on the design.)

This is an alternative to the request in #25 .

# The Short Version
Support a `service.beta.openshift.io/serving-cert-secret-name` annotation on StatefulSets, to request generating a single secret with a set of individual certificate+key pairs named `{0,1,2,3,…}.{cert,key}`. The subject of `$podID.cert` is `${statefulset.name}-${podID}.${service.name}.${service.namespace}.svc` . Each pod from a StatefulSet can then pick and use its own certificate.

Creating a StatefulSet triggers both a Secret creation and creating the StatefulSet pods, which race against each other. So, the pods **must** fail if a certificate for their identity is not found. We also generate a few more certificate+key pairs than necessary, to allow scaling up without such pod failures.

# Implementation Notes
Basically, the implementation copy & pastes the existing two Service→Secret controllers to obtain Statefulset→Secret controllers. The two are fairly orthogonal.

The PR includes is a set of small re-factoring changes, a few very small changes to the existing implementation (separate #143 being the largest one), and two commits adding the new implementation. See individual commit messages for details of the refactoring.

Some specific design/implementation concerns:

## Pod Creation Race
Unlike Services, which can be created (and trigger Secret creation) well in advance of creating Deployments etc. that rely on the secrets, creating (or enlarging) a StatefulSet triggers both creating the Secret and Pods, at the same time. So, it’s possible for a Pod to start without the Secret being available yet, or for a StatefulSet scale-up to start a Pod with a Secret that does not yet contain a certificate for the new Pod’s podID.
### Application Impact
To handle this, the application must be implemented so that the Pod fails (and triggers a retry) if the pod’s individual certificate is not available.
### Pre-creating Certificates
Secondarily, to avoid the race in predictable situations, the controller creates some extra certificates on top of the StatefulSet.spec.replicas request, so that a future small scale-up can succeed at the first try.

The implementation currently hard-codes creating 5 extra certificates, or 30% more than spec.replicas, whichever is larger.

If the StatefulSet is scaled down, we don’t immediately delete the extra certificates; that would just be extra traffic for no benefit. The secret is trimmed to a smaller size only when re-created (on certificate expiry or a resync, e.g. if the user manually deletes or corrupts the secret).
## Annotations
StatefulSets now use the same annotations as Services (serving-sert-secret-name as input, serving-cert-{created-by,generation-error,generation-error-num} as output.

Secrets can now have originating-StatefulSet-{name,uid} annotations to link back to the StatefulSet; this works just like the previous originating-service-*annotations.
### Alpha Annotations Not Used
The StatefulSet controllers only use “service.beta.openshift.io” annotations, not the “alpha” annotation which are supported for services. This is new code, so supporting an older version seems unnecessary.
## The OpenShift “Service UID” Certificate Extension Is Not Used
The service-ca certificates created for services include an OpenShift-specific “Service UID” extension, per https://github.com/openshift/origin/pull/12413 to identify intra-cluster ElasticSearch traffic.

That extension is specifically documented to contain a service UID, so it can’t be directly used for StatefulSets, and it’s not obvious that looking up the UID of a service referenced by StatefulSet.spec.serviceName makes much sense — that StatefulSet→Service reference is UID-agnostic, and it’s a different set of certificates that was not anticipated by the current consumers of the “service UID” certificate extension.

If necessary, we can allocate and define a new “StatefulSet UID” extension, or look up a Service UID for StatefulSet.spec.serviceName, or maybe even store a StatefulSet UID in the “Service UID”  extension, in the future; it seems unnecessary for a first version of this feature.
## Secret Size Grows with Replicas
Storing all certificates for a StatefulSet in a single Secret could eventually make the Secret pretty large. Hopefully that’s not a limit we are likely to hit?

It would be possible to limit the maximum Secret size (e.g. to only store up to 100? certificates in a single Secret), but the author of the StatefulSet would have to manually refer to all of these Secrets in the Pod template (and to add more references, or remove some, on scaling the StatefulSet up, or down).


mtrmac:
<summary>

A skeleton of a server that can work with this, detecting whether the right certificate is available and aborting if not.

(And also accepting both external client connections and intra-StatefulSet connections on a single port and serving the right certificate, something not necessary for the design, just to show it is possible and we don’t need a single certificate with multiple subjects matching different services.)

<details>

`test-server.go`:

```go
package main

import (
	"crypto/tls"
	"crypto/x509"
	"fmt"
	"log"
	"net/http"
	"os"
	"path/filepath"
	"strconv"
	"strings"
)

func getStatefulSetID() (int, error) {
	fullHostname, err := os.Hostname()
	if err != nil {
		return -1, fmt.Errorf("os.Hostname: %v", err)
	}
	components := strings.Split(fullHostname, ".")
	hostname := components[0]
	dash := strings.LastIndex(hostname, "-")
	if dash == -1 {
		return -1, fmt.Errorf("Unexpected host name %q: no dash in %q", fullHostname, hostname)
	}
	id, err := strconv.Atoi(hostname[dash+1:])
	if err != nil {
		return -1, fmt.Errorf("Error parsing ID in %q: %v", hostname, err)
	}
	return id, nil
}

func requestHandler(w http.ResponseWriter, r *http.Request) {
	sni := "(no SNI)"
	if r.TLS != nil {
		sni = fmt.Sprintf("SNI %#v", r.TLS.ServerName)
	}
	log.Printf("Request for %s %s", sni, r.URL.String())
	fmt.Fprintf(w, "Hello from %s %s", sni, r.URL.String())
}

func main() {
	if len(os.Args) < 3 {
		log.Fatalf("Usage: $prog listen-address dirs")
	}
	listenAddress := os.Args[1]
	certDirs := os.Args[2:]

	statefulSetID, err := getStatefulSetID()
	if err != nil {
		log.Fatalf("%v", err)
	}
	log.Printf("Host ID %d", statefulSetID)

	certs := []tls.Certificate{}
	perIDCertName := fmt.Sprintf("%d.crt", statefulSetID)
	gotPerIDCert := false
	for _, certDir := range certDirs {
		certPaths, err := filepath.Glob(filepath.Join(certDir, "*.crt"))
		if err != nil {
			log.Fatalf("glob: %v", err)
		}
		if len(certPaths) == 0 {
			log.Fatalf("No certificates found in %s", certDir)
		}
		for _, certPath := range certPaths {
			keyPath := strings.TrimSuffix(certPath, ".crt") + ".key"
			cert, err := tls.LoadX509KeyPair(certPath, keyPath)
			if err != nil {
				log.Fatalf("Error loading %s / %s: %v", certPath, keyPath, err)
			}
			cert.Leaf, err = x509.ParseCertificate(cert.Certificate[0])
			if err != nil {
				log.Fatalf("Error parsing %s / %s: %v", certPath, keyPath, err)
			}
			log.Printf("Loaded %s: %v", certPath, cert.Leaf.DNSNames)
			certs = append(certs, cert)
			if filepath.Base(certPath) == perIDCertName {
				gotPerIDCert = true
				log.Printf("\t(per-ID certificate)")
			}
		}
	}
	if !gotPerIDCert {
		log.Fatalf("No per-ID certificate found, terminating.")
	}

	server := &http.Server{
		Addr:    listenAddress,
		Handler: http.HandlerFunc(requestHandler),
		TLSConfig: &tls.Config{
			Certificates: certs,
		},
	}
	if err := server.ListenAndServeTLS("", ""); err != nil {
		log.Fatalf("Error serving: %v", err)
	}
}
```
`service.yaml`:
```yaml
kind: Service
apiVersion: v1
metadata:
  name: stateful-test
  namespace: default
  annotations:
    service.beta.openshift.io/serving-cert-secret-name: stateful-test
spec:
  ports:
    - protocol: TCP
      port: 2021
      targetPort: 2021
  selector:
    pod-label: stateful-test-pod
  clusterIP: None
```
`StatefulSet.yaml`:
```yaml
kind: StatefulSet
apiVersion: apps/v1
metadata:
  annotations:
    service.beta.openshift.io/serving-cert-secret-name: stateful-test-for-set
  name: stateful-test-set
  namespace: default
spec:
  replicas: 3
  selector:
    matchLabels:
      pod-label: stateful-test-pod
  template:
    metadata:
      labels:
        pod-label: stateful-test-pod
    spec:
      volumes:
        - name: certs
          secret:
            secretName: stateful-test
            defaultMode: 420
        - name: certs-stateful
          secret:
            secretName: stateful-test-for-set
            defaultMode: 420
      containers:
        - name: test-server
          image: 'quay.io/mitr/test-server:latest'
          command: ["/usr/bin/test-server", ":2021", "/srv/certificates", "/srv/certificates-stateful"]
          ports:
            - containerPort: 2021
              protocol: TCP
          volumeMounts:
            - name: certs
              mountPath: /srv/certificates
            - name: certs-stateful
              mountPath: /srv/certificates-stateful
  serviceName: stateful-test
```
</summary>

marun:
Has anyone talked to you about writing an enhancement for this change (https://github.com/openshift/enhancements)? While I appreciate how detailed this PR's description is, a PR comment is not a venue for discussion, and unlike the enhancement it won't outlive this PR.

pkg/controller/servingcert/controller/stateful_secret_creating_controller.go
```diff
@@ -0,0 +1,217 @@
...
+	// check the secret since it could not have been created yet
+	secretName := statefulSet.Annotations[api.ServingCertSecretAnnotation]
+	if len(secretName) == 0 {
+		return false
+	}
+
+	secret, err := sc.secretLister.Secrets(statefulSet.Namespace).Get(secretName)
+	if kapierrors.IsNotFound(err) {
+		// we have not created the secret yet
+		return true
```
mtrmac, line 146: Note to self: This ignores `getStatefulSetNumFailures` and will keep trying to create the secret (e.g. if it is too large).

mtrmac: Fixed.

mtrmac:
(Updated to use `.alpha.` annotations instead of `.beta.`, based on the discussion in https://github.com/openshift/enhancements/pull/674 . Still needs e2e tests.)

mtrmac:
This is now as complete as I can make it, pending decisions in https://github.com/openshift/enhancements/pull/674 . Lots of small refactoring + the truly new code in the there `WIP:` commits.

Note that this PR includes (as branch merges) currently unmerged #143 and #147 .

PR #147, opened 2021-03-09
Various e2e test fixes and cleanups

mtrmac:
This is a set of fixes to e2e tests
- fixing some bugs (or short timeouts) that show up as flakes (e.g. in https://prow.ci.openshift.org/pr-history/?org=openshift&repo=service-ca-operator&pr=143 )
- improving the code to make test development easier
- a few minor cleanups.

See individual commit messages for details.

Note primarily the commit “Don't abort metrics/service-ca-metrics before the sync finishes”, where I could see an argument that this should be changed in the code instead of the tests.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -58,7 +58,7 @@ const (
 	// Rotation of all certs and bundles is expected to take a considerable amount of time
 	// due to the operator having to restart each controller and then each controller having
 	// to acquire the leader election lease and update all targeted resources.
-	rotationTimeout = 3 * time.Minute
+	rotationTimeout = 5 * time.Minute
 	// Polling for resources related to rotation may be delayed by the number of resources
 	// that are updated in the cluster in response to rotation.
 	rotationPollTimeout = 2 * time.Minute
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -100,7 +100,6 @@ func createTestNamespace(client *kubernetes.Clientset, namespaceName string) (*v
 	return ns, err
 }
 
-// on success returns serviceName, secretName, nil
 func createServingCertAnnotatedService(client *kubernetes.Clientset, secretName, serviceName, namespace string) error {
 	_, err := client.CoreV1().Services(namespace).Create(context.TODO(), &v1.Service{
 		TypeMeta: metav1.TypeMeta{},
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -170,8 +170,8 @@ func editServingSecretData(client *kubernetes.Clientset, secretName, namespace, 
 	if err != nil {
 		return err
 	}
-	time.Sleep(10 * time.Second)
-	return nil
+
+	return pollForSecretChange(client, scopy, keyName)
 }
 
 func editConfigMapCABundleInjectionData(client *kubernetes.Clientset, configMapName, namespace string) error {
@@ -188,8 +188,8 @@ func editConfigMapCABundleInjectionData(client *kubernetes.Clientset, configMapN
 	if err != nil {
 		return err
 	}
-	time.Sleep(10 * time.Second)
-	return nil
+
+	return pollForConfigMapChange(client, cmcopy, "foo")
 }
 
 func checkServiceServingCertSecretData(client *kubernetes.Clientset, secretName, namespace string) ([]byte, bool, error) {
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -205,21 +205,23 @@ func checkServiceServingCertSecretData(client *kubernetes.Clientset, secretName,
 	if len(sss.Data) != 2 {
 		return nil, false, fmt.Errorf("unexpected service serving secret data map length: %v", len(sss.Data))
 	}
-	ok := true
-	_, ok = sss.Data[v1.TLSCertKey]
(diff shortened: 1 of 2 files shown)


mtrmac:
As an aside: To determine the need to increase timeouts  I have added simple logging of time spent polling in the various larger-timeout cases (around the `wait.PollImmediate` calls, basically), and that code did appear to work fine (e.g. on failures the reported times match the timeouts, and the further indirectly-caused failures also show plausible time changes).

The curious thing is that with the test runs that do fully succeed, the reported polling times match _exactly_, down to the nanosecond, across separate test runs on the same cluster. On a rebuilt cluster the times are different from the first cluster, but again all of them match _exactly_.

Some part of that can probably be explained by low resolution of the macOS clock, still, the consistent timing of cluster’s operation is so spooky that it’s very tempting to think I’m missing something.

test/e2e/e2e_test.go
```diff
@@ -461,7 +459,7 @@ func triggerForcedRotation(t *testing.T, client *kubernetes.Clientset, config *r
 
 	// Trigger a forced rotation by updating the operator config
 	// with a reason.
-	setUnsupportedServiceCAConfig(t, config, "42", customDuration)
+	forceUnsupportedServiceCAConfigRotation(t, config, secret, customDuration)
```
marun, line 462: (No action required) Not sure what this change is buying given this function is only called once in a test run. YAGNI and all.

test/e2e/e2e_test.go
```diff
@@ -511,6 +518,21 @@ func pollForCARotation(t *testing.T, client *kubernetes.Clientset, caCertPEM, ca
...
+// pollForCARecreation polls for the signing secret to be re-created in
+// response to CA secret deletion.
+func pollForCARecreation(client *kubernetes.Clientset) error {
+	return wait.PollImmediate(time.Second, rotationPollTimeout, func() (bool, error) {
+		_, err := client.CoreV1().Secrets(serviceCAControllerNamespace).Get(context.TODO(), signingKeySecretName, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			return false, nil
+		}
+		if err != nil {
+			return false, err
```
marun, line 530: (No action required) Classic polling mistake - returning an error exits rather than triggering retry as intended. No blame intended - lots of examples of this in this file. Best practice for handling potentially recoverable failure in a polling function is to log the error and not return it.

marun (commented):
@stlaz lgtm, feel free to approve.

test/e2e/e2e_test.go
mtrmac: This is the “Fix failures on repeated e2e tests in the same cluster” commit. 

The current behavior is perfectly fine for a CI run; when developing the tests, repeatedly running them on a single cluster, this failure is distracting from any real failures or test bugs.

test/e2e/e2e_test.go
mtrmac: Thanks, I was treating the server as “stable/available by assumption” and I didn’t realize that’s a bad assumption for HA code.

I’m not going to change this one example as an exception, but if you’d prefer me to change all of them, I can do that.

test/e2e/e2e_test.go
```diff
@@ -486,6 +486,15 @@ func setUnsupportedServiceCAConfig(t *testing.T, config *rest.Config, forceRotat
...
 	}
+	var forceRotationReason string
+	i := 0
+	for {
+		forceRotationReason = fmt.Sprintf("service-ca-e2e-force-rotation-reason-%d", i)
+		if currentSigningKeySecret.Annotations[api.ForcedRotationReasonAnnotationName] != forceRotationReason {
+			break
+		}
+		i++
+	}
```
stlaz, line 494: just a little bit of syntax sugar sprinkles, also limiting the scope of `i`
```
var forceRotationReason string
for i := 0; ; i++ {
    forceRotationReason = fmt.Sprintf("service-ca-e2e-force-rotation-reason-%d", i)
    if currentSigningKeySecret.Annotations[api.ForcedRotationReasonAnnotationName] == forceRotationReason {
        break
    }
}
```

test/e2e/e2e_test.go
```diff
@@ -521,6 +521,21 @@ func pollForCARotation(t *testing.T, client *kubernetes.Clientset, caCertPEM, ca
...
+// pollForCARecreation polls for the signing secret to be re-created in
+// response to CA secret deletion.
+func pollForCARecreation(client *kubernetes.Clientset) error {
+	return wait.PollImmediate(time.Second, rotationPollTimeout, func() (bool, error) {
+		_, err := client.CoreV1().Secrets(serviceCAControllerNamespace).Get(context.TODO(), signingKeySecretName, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			return false, nil
+		}
+		if err != nil {
+			return false, err
```
stlaz, line 530: (no action required) I usually prefer
```
if err != nil {
    if errors.IsNotFound(err) {
        return false, nil
    }
    return false, err
}
```
but I see that that's not the pattern used in this file either :neutral_face:

stlaz (commented):
This PR sums up many sins of the e2e tests in this repo and I'll gladly see it merged.

I only have a single nit about syntax sugar we can use that I'd prefer to be included and I'll bless the PR.

test/e2e/e2e_test.go
mtrmac: Thanks, updated.

stlaz:
/lgtm
thanks for the fixes \o/

PR #151, opened 2021-03-30
IBM Cloud manifest profile patch

csrwng:
Bumps build-machinery-go to latest
Adds patch specific to ibm-cloud-managed profile that removes node selector for operator deployment
make verify verifies that the generated manifest is up to date
make update updates the manifest by applying the ibm-cloud-managed patch to the original deployment manifest

This is needed for the IBM Cloud managed profile because in that profile there are no master nodes.

### DIFF
--- a/manifests/05_deploy-ibm-cloud-managed.yaml
+++ b/manifests/05_deploy-ibm-cloud-managed.yaml
@@ -1,0 +1,71 @@
+# *** AUTOMATICALLY GENERATED FILE - DO NOT EDIT ***
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+  labels:
+    app: service-ca-operator
+  name: service-ca-operator
+  namespace: openshift-service-ca-operator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: service-ca-operator
+  template:
+    metadata:
+      labels:
+        app: service-ca-operator
+      name: service-ca-operator
+    spec:
+      containers:
+      - args:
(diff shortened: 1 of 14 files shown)


manifests/05_deploy-ibm-cloud-managed.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+  labels:
+    app: service-ca-operator
+  name: service-ca-operator
```
sttts, line 8: what is this file? Why is it called ...-ibm-cloud-managed...

csrwng: It's a generated file (from 05_deploy.yaml) specific to the `ibm-cloud-managed` profile. The file is created by applying the patch in `profile-patches/ibm-cloud-managed/05_deploy.yaml-patch`. 
`make verify` ensures that the file is up to date (by applying the patch)
`make update` updates the file when either the patch or the source manifest changes.

csrwng: The reason for this, is that for any CVO-delivered deployment that will be applied in a ROKS cluster (or for that matter any hosted control plane deployment of openshift), we need to remove the node selector for masters, since in that environment there are no master machines.

sttts: I see. So when we change the main file without the patch, `make verify` will complain.

csrwng: Yes

PR #167, opened 2021-07-09
bug 1981498: add vulnerable legacy injector to allow for upgrade clusters to use ...

deads2k:
the old service-ca.crt content in SA tokens

This is part 1 of a four part plan.

1. add the ability to inject *only* an openshift-service-ca.crt configmap with content that in 4.7 was contained in token SA service-ca.crt.  It uses the `service.alpha.openshift.io/inject-vulnerable-legacy-cabundle` annotation so that it is clearly undesireable and the behavior isn't guaranteed.  The controller here yields to the preferred controller if both annotations are set.
2. add an API field to the kcm operator resource so a user can indicate they wish to switch to a more secure service-ca.crt.
3. Add a flag to the KCM to indicate "use the vulnerable annotation" during configmap creation. The bootstrap kcm does not need this since it is only used in new clusters.
4. set the new API field into the kcm-o "create-only" operator resource.  This will let new clusters start secure.  The operator will use the field to set the KCM flag

### DIFF
--- a/pkg/controller/cabundleinjector/starter.go
+++ b/pkg/controller/cabundleinjector/starter.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"os"
 	"strings"
 	"time"
 
@@ -25,6 +26,13 @@ type caBundleInjectorConfig struct {
 	caBundle      []byte
 	kubeClient    *kubernetes.Clientset
 	kubeInformers kubeinformers.SharedInformerFactory
+
+	// legacyVulnerableCABundle is a CA bundle which included more certificates than are needed for verifying service
+	// serving certificates.  This was addressed in new installs of 4.8, but migrated clusters continue to have the old
+	// content inside of their bound tokens for the service-ca.crt.
+	// This CA bundle should only be used for specifically named configmaps which explicitly indicate their desire.
+	// This makes it impossible for customers to use and being to rely upon.
+	legacyVulnerableCABundle []byte
 }
 
 type startInformersFunc func(stopChan <-chan struct{})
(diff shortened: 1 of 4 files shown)


deads2k:
the lack of unit tests is going to be sticky. :(

I guess we'll use the e2e-aws run.

```
=== RUN   TestE2E/vulnerable-legacy-ca-bundle-injection-configmap
    e2e_test.go:1376: configmaps "openshift-service-ca.crt" already exists
```

mfojtik:
/lgtm


/hold
(for Standa ack)

pkg/controller/cabundleinjector/starter.go
```diff
@@ -44,21 +52,36 @@ func StartCABundleInjector(ctx context.Context, controllerContext *controllercmd
...
 	// depending on the operator for redeployment
 	caBundleFile := "/var/run/configmaps/signing-cabundle/ca-bundle.crt"
-
 	caBundleContent, err := ioutil.ReadFile(caBundleFile)
 	if err != nil {
 		return err
 	}
 
+	// this construction matches what the old kube controller manager did. It added the entire ca.crt to the service-ca.crt.
+	vulnerableLegacyCABundleContent, err := ioutil.ReadFile(caBundleFile)
```
stlaz, line 61: nit: this re-reads the file

stlaz:
/hold cancel
looks ok

deads2k:
reapplying lgtm

test/e2e/e2e_test.go
```diff
@@ -1317,6 +1338,90 @@ func TestE2E(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		// the content should now change pretty quick.  We sleep because it's easier than writing a new poll and I'm pressed for time
+		time.Sleep(5 * time.Second)
+		publishedConfigMap, err = adminClient.CoreV1().ConfigMaps(ns.Name).Get(context.TODO(), publishedConfigMap.Name, metav1.GetOptions{})
+
+		// if we changed the injection, we should see different content
```
marun, line 1413: nit: Maybe move this comment to the second block or remove it entirely? It doesn't appear relevant to the contents of the first block, since in that case different content is not expected.

marun (commented):
Some nits...

PR #196, opened 2022-07-01
Bug 2048349: make the operator react to workload logLevel configuration

stlaz:
The "ApplyWithForce" logic does not react to changing the Pod
template spec.

/assing @ibihim


stlaz:
/test all
the tests should now be fixed

PR #202, opened 2022-11-03
Return nil from controller start func after context is cancelled.

benluddy:
Returning an error here triggers leader election error handling codepaths and prevents the release of the leader lease during graceful shutdown.


stlaz:
/hold
there are more places like this

mangelajo:
nice @benluddy thank you for handling this so quickly

mangelajo:
> /hold there are more places like this
> 
> edit: the only other place I found is
> 
> https://github.com/openshift/service-ca-operator/blob/42a7a275009f8c534fe92c6ba577d168a549eadd/pkg/operator/starter.go#L136-L137

it seems to be handled now

/unhold

PR #263, opened 2025-04-16
OCPBUGS-55057: Don't check featuregates on operand

vrutkovs:
Microshift doesn't have FeatureGates CRD, so operand should not attempt to read featuregates directly and instead the operator should pass featuregates as cli argument.


pkg/controller/servingcert/starter/starter.go
```diff
@@ -50,33 +53,36 @@ func StartServiceServingCertSigner(ctx context.Context, controllerContext *contr
...
-		klog.Errorf("timed out waiting for FeatureGate detection")
-		return fmt.Errorf("timed out waiting for FeatureGate detection")
-	}
 
 	minTimeLeftForCert := time.Hour
 	certificateLifetime := 2 * 365 * 24 * time.Hour
-	if featureGates.Enabled(features.FeatureShortCertRotation) {
-		minTimeLeftForCert = time.Hour
-		certificateLifetime = time.Hour * 2
+	if ok, err := IsMicroshift(ctx, kubeClient); err == nil && !ok {
```
benluddy, line 61: Can we avoid creating a precedent for adding MicroShift special cases to OCP components?

The cluster-config-operator includes code to render configv1 FeatureGate manifest files. Could we instead add an option to service-ca-operator to provide a FeatureGate manifest to be used instead of loading them from the API? MicroShift would be able to configure that option appropriately, and service-ca-operator would not need to detect or even be aware of MicroShift.

vrutkovs: @pmtk wdyt?

pkg/controller/servingcert/starter/starter.go
```diff
@@ -139,3 +145,14 @@ func readIntermediateCACert(filename string) (*x509.Certificate, error) {
 	}
 	return certs[0], nil
 }
+
+// IsMicroshift checks if the operator is running on Microshift.
+func IsMicroshift(ctx context.Context, kubeClient *kubernetes.Clientset) (bool, error) {
+	if _, err := kubeClient.CoreV1().ConfigMaps("kube-public").Get(ctx, "microshift-version", metav1.GetOptions{}); err != nil {
```
sjenning, line 151: Is this really the proper way to determine if a cluster is a microshift cluster?

pkg/operator/starter.go
```diff
@@ -150,9 +134,30 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	// trusted) should be valid for at least this long.
 	minimumTrustDuration := 395 * 24 * time.Hour
 
-	if featureGates.Enabled(features.FeatureShortCertRotation) {
-		minimumTrustDuration = time.Hour + 15*time.Minute
-		signingCertificateLifetime = time.Hour*2 + 30*time.Minute
+	if ok, err := starter.IsMicroshift(ctx, kubeClient); err == nil && !ok {
```
sjenning, line 137: I don't think we need to detect Microshift in the operator because Microshift does not run the operator, just the operand.

sjenning (changes requested):

pkg/controller/servingcert/starter/starter.go
sjenning: These are the components run by Microshift, among them `service-ca` (the operand, not the operator)

https://github.com/openshift/microshift/tree/main/assets/components

Afaict, the only other controller that Microshift runs that is feature gate aware is the `csi-snapshot-controller`, which takes gates as a flag
https://github.com/openshift/csi-external-snapshotter/blob/ac82cafc95b301f67f46ee0db93720d55177a19b/cmd/snapshot-controller/main.go#L154

Microshift just doesn't set the flag.

Looking back at https://github.com/openshift/service-ca-operator/pull/245, we should have added the FeatureGate informer to the operator (like we did), but added a `feature-gates` flag to the controller rather than have it directly observe the FeatureGate config.

pmtk: I think that makes sense, it would be more elegant solution

pkg/cmd/controller/cmd.go
```diff
@@ -10,11 +13,25 @@ import (
...
+)
+
 func NewController() *cobra.Command {
 	cmd := controllercmd.
-		NewControllerCommandConfig("service-ca-controller", version.Get(), controller.StartServiceCAControllers, clock.RealClock{}).
+		NewControllerCommandConfig("service-ca-controller", version.Get(), startServiceCAControllers, clock.RealClock{}).
 		NewCommand()
 	cmd.Use = "controller"
 	cmd.Short = "Start the Service CA controllers"
+	cmd.Flags().Var(utilflag.NewMapStringBool(&featureGates), "feature-gates", "Comma-separated list of key=value pairs that describe feature gates for alpha/experimental features.")
```
benluddy, line 26: Did you consider creating a component-base feature gate object to hold these? The component-base type includes an `AddFlag(*pflag.FlagSet)` method for this purpose, and it would behave exactly the same as the --feature-gates option on kube components.

benluddy: Seems OK in the context of fixing the bug.

> This might be a good idea for overall cmd flag standartization across the whole platform though

Having the operator wire them to a --feature-gates flag is our convention for at least OCM, KAS, KCM, and OAS. I think it's a good idea to stick to that in general. Making the mechanism consistent encourages the use of feature gates across the platform.

pkg/cmd/controller/cmd.go
```diff
@@ -10,11 +13,25 @@ import (
 	"github.com/openshift/service-ca-operator/pkg/version"
 )
 
+var (
+	featureGates map[string]bool
```
benluddy, line 17: I'm a little concerned about future changes creating the potential for concurrent reads and writes of this map. Maybe you could instead declare this map within the call to `NewController` and replace `startServiceCAControllers` with a closure capturing it, rather than sharing state through this global?

vrutkovs: Good idea, reworked this with a local var and closure

pkg/cmd/controller/cmd.go
```diff
@@ -11,10 +14,20 @@ import (
...
+	var featureGates map[string]bool
+
 	cmd := controllercmd.
-		NewControllerCommandConfig("service-ca-controller", version.Get(), controller.StartServiceCAControllers, clock.RealClock{}).
+		NewControllerCommandConfig("service-ca-controller", version.Get(), func(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
+			shortCertRotationEnabled := false
+			if val, ok := featureGates["ShortCertRotation"]; ok && val {
+				shortCertRotationEnabled = true
+			}
+			return controller.StartServiceCAControllers(ctx, controllerContext, shortCertRotationEnabled)
```
benluddy, line 25: Nit: This logic is unnecessary since it matches the behavior of `shortCertRotationEnabled := featureGates["ShortCertRotation"]`.

```suggestion
			return controller.StartServiceCAControllers(ctx, controllerContext, featureGates["ShortCertRotation"])
```

vrutkovs: Oh indeed, updated

PR #278, opened 2025-10-08
CNTRLPLANE-1544: controller deployment: Use user namespace

tchap:
Switch to restricted-v3 scc and enable user namespaces.

### DIFF
--- a/bindata/v4.0.0/controller/deployment.yaml
+++ b/bindata/v4.0.0/controller/deployment.yaml
@@ -18,22 +18,20 @@ spec:
       name: service-ca
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
-        openshift.io/required-scc: restricted-v2
+        openshift.io/required-scc: restricted-v3
       labels:
         app: service-ca
         service-ca: "true"
     spec:
       serviceAccountName: service-ca
+      hostUsers: false
       containers:
       - name: service-ca-controller
         image: ${IMAGE}
         imagePullPolicy: IfNotPresent
-        terminationMessagePolicy: FallbackTo
         command: ["service-ca-operator", "controller"]
         ports:
         - containerPort: 8443
-        securityContext:
-          runAsNonRoot: true
         resources:
           requests:
             memory: 120Mi
--- a/pkg/operator/v4_00_assets/bindata.go
+++ b/pkg/operator/v4_00_assets/bindata.go
@@ -207,22 +207,20 @@ spec:
       name: service-ca
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
-        openshift.io/required-scc: restricted-v2
+        openshift.io/required-scc: restricted-v3
       labels:
         app: service-ca
         service-ca: "true"
     spec:
       serviceAccountName: service-ca
+      hostUsers: false
       containers:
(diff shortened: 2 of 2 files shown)


bindata/v4.0.0/controller/deployment.yaml
```diff
@@ -18,27 +18,33 @@ spec:
...
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
-        openshift.io/required-scc: restricted-v2
+        openshift.io/required-scc: restricted-v3
       labels:
         app: service-ca
         service-ca: "true"
     spec:
       serviceAccountName: service-ca
+      hostUsers: false
```
p0lyn0mial: ok, i understand that this is required to opt-in into user namespaces

bindata/v4.0.0/controller/deployment.yaml
```diff
@@ -18,27 +18,33 @@ spec:
       name: service-ca
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
-        openshift.io/required-scc: restricted-v2
+        openshift.io/required-scc: restricted-v3
```
p0lyn0mial: but why do we need to set this ?

p0lyn0mial: but why do we want to use that policy ?

tchap: Actually now that I think about it, it's required for user namespaces. Because it explicitly sets the allowed UID range to the one required by user namespaces. Otherwise the UID range is picked up based on the annotation on the associated namespace and that cannot be used for user namespaces. So this is why: 

https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-restricted-v3.yaml#L40

bindata/v4.0.0/controller/deployment.yaml
```diff
@@ -18,27 +18,33 @@ spec:
...
       labels:
         app: service-ca
         service-ca: "true"
     spec:
       serviceAccountName: service-ca
+      hostUsers: false
+      securityContext:
```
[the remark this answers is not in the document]
tchap: Let me check whether this needs to be set explicitly. It's required by the scc policy, but it may be set automatically so that it doesn't have to be present in the file really.

bindata/v4.0.0/controller/deployment.yaml
tchap: Yeah, exactly. It implies some defaults, but also enforces the spec is not in conflict with the policy.

bindata/v4.0.0/controller/deployment.yaml
```diff
@@ -18,27 +18,33 @@ spec:
...
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - name: service-ca-controller
         image: ${IMAGE}
         imagePullPolicy: IfNotPresent
-        terminationMessagePolicy: FallbackTo
```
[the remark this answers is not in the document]
tchap: It is already set to a correct value on line 42. This line is extra.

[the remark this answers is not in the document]
p0lyn0mial: maybe that's why it worked :)

PR #312, opened 2026-02-03
Fix for OCPBUGS-67872: CVE-2025-65637 bump logrus to github.com/sirupsen/logrus v1.9.3 

germanparente:
bump logrus to github.com/sirupsen/logrus v1.9.3


germanparente:
After checking more deeply, the vulnerable symbols are not used in the component. Closing PR.

PR #319, opened 2026-02-09
MON-4515: Migrate Prometheus targets discovering from Endpoints to EndpointSlices

machine424:
This PR migrates Prometheus service discovery from the deprecated Endpoints API to the EndpointSlices API, by:

- Setting `serviceDiscoveryRole: EndpointSlice` on ServiceMonitors.
- Granting Prometheus `endpointslices` permissions.

We're taking a conservative approach by keeping the existing `endpoints` permissions alongside the new `endpointslices` ones. This provides a safety net in case any ServiceMonitors, whether deployed from this repo or from another source, still rely on the same Role and were missed during the migration.

That said, since both resources provide essentially the same data, keeping both isn't meaningfully more permissive from a security standpoint.

**These changes target OpenShift 4.22+ and should not be backported to earlier releases.**

Due to the scope of changes across multiple repositories, these modifications were generated with Claude assistance.

### DIFF
--- a/manifests/0000_90_service-ca-operator_01_prometheusrole.yaml
+++ b/manifests/0000_90_service-ca-operator_01_prometheusrole.yaml
@@ -19,3 +19,11 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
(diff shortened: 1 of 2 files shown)


manifests/0000_90_service-ca-operator_01_prometheusrole.yaml
everettraven, line 16: Do we need to maintain the permissions on the `endpoints` resource as part of this migration?

machine424: Please check https://github.com/openshift/service-ca-operator/pull/319#issue-3915881677

PR #323, opened 2026-03-03
NO-JIRA: controller/servingcerts: Align log verbosity

tchap:
Some informational statements are being logged on V(4), which makes then effectively invisible in production clusters. It would be handly to have them for analyzing cluster issues, though.

## Summary by CodeRabbit

* **Chores**
  * Improved logging visibility for operational diagnostics. Key system messages are now more readily available in standard debug output, enhancing observability and simplifying troubleshooting without requiring verbose logging configurations.


sanchezl (approved):
/lgtm

tchap:
/verified bypass

There is nothing to test really.

PR #329, opened 2026-03-13
OCPBUGS-73938: Watch CA bundle files for changes and reload dynamically

tchap:
Replace the static one-shot file read at startup with a bundlesWatcher
that monitors the CA bundle and SA token CA files using
filesystem.WatchUntil. When either file changes on disk, the watcher
reloads the affected cache and rebuilds the legacy bundle. Injector
controllers now read from the shared bundleCache instead of holding a
static []byte, so CA rotations take effect without a pod restart.

### For the Reviewer

We made the decision not to update all configmaps, as much as that would be more clear. This PR is to handle a rather obscure client case, as far as I remember. It can be changes if deemed necessary.

## Summary by CodeRabbit

* **New Features**
  * CA bundles are now watched and stored in an atomic cache with automatic reloads on file changes.

* **Improvements**
  * Injection now loads CA data at sync time (removes synchronous file reads at startup) and uses the runtime cache, improving reliability.

* **Bug Fixes**
  * Syncs fail early with an error when CA data is missing.

* **Tests**
  * Added comprehensive tests for watcher reads, reloads, observer events, resync behavior, and error scenarios.


pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	observer.AddReactor(func(_ string, _ fileobserver.ActionType) error {
+		return w.reloadBundles(ctx)
+	}, map[string][]byte{
+		w.caBundlePath:        caBundleData,
+		w.saTokenCABundlePath: saTokenCAData,
+	}, w.caBundlePath, w.saTokenCABundlePath)
+
+	go observer.Run(ctx.Done())
```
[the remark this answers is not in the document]
tchap: We explicitly decided not to do this, because that would require more changes, but it could be added for sure. This behavior is rather obscure and it attends to the original client case.

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (w *bundlesWatcher) readBundles() ([]byte, []byte, []byte, error) {
+	caBundleContent, err := w.readFileWithRetries(w.caBundlePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil, nil, nil
+		}
+		return nil, nil, nil, fmt.Errorf("failed to read %q: %w", w.caBundlePath, err)
+	}
```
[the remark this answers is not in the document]
tchap: I think that retaining the bundle even though it's been deleted from the disk is not a good approach as deletion means bundles are not available any more and we should not use them. Feel somehow insecure.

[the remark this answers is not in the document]
tchap: Not really, that is an outdated description.

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+	return nil
+}
+
+func (c *bundleCache) Store(data []byte) {
+	if data == nil {
+		c.v.Store(nil)
+	} else {
+		c.v.Store(&data)
+	}
```
everettraven: Nit: It is more idiomatic Go to use early returns instead of `if-else` blocks.

```suggestion
        return
	}
		
	c.v.Store(&data)
```

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+	// caBundle contains the signing CA bundle.
+	CABundle *bundleCache
+	// caBundleLegacy is constructed so that it matches what the old KCM used to do.
+	// The signing CA bundle is appended to the service account ca.crt.
+	CABundleLegacy *bundleCache
+
+	// OnReadFailed can be set for custom read error handling.
+	// When the hook returns the original error, it is logged.
+	// Logging happens automatically when no hook is set.
+	OnReadFailed func(context.Context, error) error
```
everettraven: What is the use case of this field?

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+
+	caBundlePath        string
+	saTokenCABundlePath string
+	pollingInterval     time.Duration
+	resyncInterval      time.Duration
+	maxReadAttempts     int
+
+	// readFile is used to read files from disk.
+	// Used for mocking during tests mostly.
+	readFile func(string) ([]byte, error)
```
everettraven, line 51: We could probably use something like https://pkg.go.dev/k8s.io/kubernetes/pkg/util/filesystem#Filesystem in place of this override function so that we have a more standard filesystem abstraction

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+		return w.reloadBundles(ctx)
+	}, map[string][]byte{
+		w.caBundlePath:        caBundleData,
+		w.saTokenCABundlePath: saTokenCAData,
+	}, w.caBundlePath, w.saTokenCABundlePath)
+
+	go observer.Run(ctx.Done())
+
+	// Periodically re-read the bundles as a safety net
+	// in case the file observer reactor callback fails.
```
everettraven: Under which circumstances might the file observer reactor callback fail in a way that means we need this periodic re-sync?

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+	w.CABundleLegacy.Store(legacyCABundleData)
+
+	// Start watching.
+	observer, err := fileobserver.NewObserver(w.pollingInterval)
+	if err != nil {
+		return fmt.Errorf("failed to start CA bundles observer: %w", err)
+	}
+
+	observer.AddReactor(func(_ string, _ fileobserver.ActionType) error {
+		return w.reloadBundles(ctx)
```
everettraven: Do we actually want this to happen for _any_ action type? How does this handle the file being removed?

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	var lastErr error
+	for range numAttempts {
+		content, err := w.readFile(path)
+		if err == nil {
+			return content, nil
+		}
+		lastErr = err
+	}
```
everettraven: Why do we need to retry the file reads?

pkg/controller/cabundleinjector/bundles_watcher.go
everettraven: Update: I saw this ended up being used to record an event on a failure to reload the bundles - could just pass the event recorder to this.

pkg/controller/cabundleinjector/bundles_watcher.go
everettraven: Additionally, should we perform a more nuanced reload based on the file that was changed?

Both bundle caches don't need to be refreshed if only the SA CA file changes.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: We only ever need to read a file, so this seemed like the simplest solution, allowing us to mock errors or any calls in the tests really without needing to write any files to the disk. Using a full interface and then mocking a single call seems like a bit of an overkill to me.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: `readBundles` down in the call stack actually returns `nil` data on not found, so this call handles all possible action types just fine. It just sets the cache data to `nil` on file removed.

tchap: I am wondering whether the more nuanced way is worth the complexity. I don't think these files change too often, right? Until now the change was never handled and it kinda worked for most people 😅 But in principle sure, we could read the content from cache for the file that is unchanged.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: The issue with the observer is that it stores the current file hash before invoking the callback, so if you fail in the callback, it is never retried since the current file hash matches the hash of the file on disk. I personally find this really weird, but it is what it is. Hence the resync loop.

I don't think it is safe to ignore the read error in any case, whatever the reason may be. There is simply a possibility to never update the file on read error, which needs to be avoided.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: I probably forgot this in place after I added the resync loop, will remove this.

tchap: Removed in e10b49ff6

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: Aligned in ac9e902a1

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: Amended in 030081c66

tchap:
Got a data race in tests, blocked by https://github.com/openshift/library-go/pull/2261 now.

/hold

tchap:
Updated library-go, the data race in tests should be gone now.

/unhold

tchap:
Let me actually

/hold

this so that I can make sure the commits are squashed before merging.

pkg/controller/cabundleinjector/bundles_watcher.go
everettraven: Not strongly opinionated here so I'm fine with leaving this as-is.

Just thought I'd share that there is a kube util package with a filesystem abstraction already.

pkg/controller/cabundleinjector/bundles_watcher.go
everettraven: Would using something like https://pkg.go.dev/k8s.io/apiserver@v0.36.1/pkg/util/filesystem#WatchUntil help remediate the issues there?

It looks like it handles polling as well as periodically re-syncing the file watch as needed depending on the operation that has occurred.

pkg/controller/cabundleinjector/bundles_watcher.go
everettraven: If the file is removed and we end up setting the cache data to `nil`, won't that cause every sync operation on the controllers that rely on this data to return an error when attempting to load the ca bundle from the cache? Do we actually want that?

The alternatives seem like they would be:
- we treat `nil` as "empty" and the sync functions just overwrite existing ca bundles with the empty data.
- we treat `nil` as "empty" and the sync functions do nothing (i.e nothing to sync). This would leave existing ca bundles as-is which may eventually become stale.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: Yeah, we could use this, thanks. There are too many file watching packages 🥲

tchap: Hmm, in 1.35 this is available only in `k8s.io/kubernetes/pkg/util/filesystem`. And it was moved to the given package in 1.36, it seems, which means it will explode on the next rebase... I can still use this and add a comment to change the import path in the next release, provided we are ok with importing `kubernetes` now.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: The thing is, both files being watched should always exist. They are mounted as volumes and the operator ensures the bundle ConfigMap is created before the deployment: https://github.com/openshift/service-ca-operator/blob/914254b51a328935da3c7ea62a550e3f33727d41/pkg/operator/sync.go#L31

Previously the whole executable simply crashed on the files not being there, so I just changed this to start failing in `sync`.

tchap: Hmm, I think that crashing initially still when the bundle is not there makes sense, so I added it in 5984beedd22cb5c2afef336f653a42d881a68a13 . Let me know what you think...

pkg/controller/cabundleinjector/bundles_watcher.go
everettraven: I think that should be fine

pkg/controller/cabundleinjector/bundles_watcher.go
everettraven: If that is how it was, I'm fine with keeping it that way.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: I occurred to me `WatchUntil` only handles a single path, so we would need to run it twice and synchronize. It can be done no prob, but it seems arguably more complicated than what is already implemented. Let me know what you think.

everettraven: Two function calls to a battle-tested implementation seems less complicated than hand-rolling a periodic re-sync implementation to me.

If you feel strongly that the current approach is less complicated than importing the same file watching utility that many of the core Kubernetes components use, I won't block on using the import.

tchap: I changed the implementation as you proposed, feel free to check ce604c841737993be33881260d0147834f57aeae . The reload logic is basically expecting the files to always be there, so not found is treated as any other error. There is also some debounce logic, I am wondering whether it is worth the complexity, but otherwise we are always reloading twice on every change.

everettraven: >There is also some debounce logic, I am wondering whether it is worth the complexity, but otherwise we are always reloading twice on every change

Where are you seeing the second reload on every change? The underlying library looks like it only loads once on an event to the file?

It does look like it loads once on startup though, so doing the initial reload outside of the `WatchUntil` call looks like it would make it look like we are doing another reload.

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,146 @@
...
+		coalesceDuration:    defaultCoalesceDuration,
+		recorder:            recorder,
+		readFile:            os.ReadFile,
+	}
+}
+
+func (w *bundlesWatcher) Start(ctx context.Context) error {
+	if err := w.reloadBundles(); err != nil {
+		return err
+	}
```
everettraven: Doing this initial call shouldn't be necessary outside of the `WatchUntil` calls below. Those calls should call the event handler on startup and then it is called on any subsequent events.

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,146 @@
...
+		default:
+		}
+	}
+	watchErrorHandler := func(path string) func(error) {
+		return func(err error) {
+			klog.Errorf("Error watching %q: %v", path, err)
+		}
+	}
+	go filesystem.WatchUntil(ctx, w.pollingInterval, w.caBundlePath, notify, watchErrorHandler(w.caBundlePath))
+	go filesystem.WatchUntil(ctx, w.pollingInterval, w.saTokenCABundlePath, notify, watchErrorHandler(w.saTokenCABundlePath))
```
everettraven: Presumably the debouncing logic is necessary because these both call the same reload behavior on these file changes - and they are very likely to end up changing at roughly the same time so we need to de-duplicate those?

It seems like having each of these load only the data they care about on a change would eliminate the need for the debouncing logic.

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: They should, but also they run in a goroutine, so it feels safer to make sure the caches are populated before we return from `Start`. I think that it would ok since eventual consistency, but I am always more happy to have less of that and less of log spam 😅

pkg/controller/cabundleinjector/bundles_watcher.go
tchap: Yeah, exactly. I actually drafted it initially the way you mention, but the update functions were terribly unreadable compared to this. But I will try one more time to get it to a readable shape.

tchap: Ok, not sure what I was doing last time, it's actually ok: 08ed943c3a144e1bcd4e3f81391b56ead0dacecd

pkg/controller/cabundleinjector/bundles_watcher.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (w *bundlesWatcher) reloadSATokenCALogged() {
+	if err := w.reloadSATokenCA(); err != nil {
+		klog.Errorf("Failed to reload SA token CA bundle: %v", err)
+		w.recorder.Warningf("CABundleReloadFailed", "Failed to reload SA token CA bundle: %v", err)
+	}
+}
+
+func (w *bundlesWatcher) rebuildLegacyBundle() {
```
everettraven, line 128: I've not used the `atomic` library much and I'm drawing a bit of a blank here - do we need to have any kind of thread safety in place here since this ends up being called by both watchers?

tchap: We are good. Store/Load is atomic and we call this function on change in any of the files. There is no locking, so you could get calls that are interleaved in a way so that you get a weird legacy bundle, but the bundle is updated as the last call on any change, so this always end up being consistent.

tchap:
Commits squashed

/unhold

PR #338, opened 2026-03-25
CNTRLPLANE-2894:e2e network policy test cases

kaleemsiddiqu:
Network policy tests for added network policy for service-ca-operator component

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -26,7 +26,6 @@ import (
 )
 
 const (
-	serviceCAOperatorNamespace   = operatorclient.OperatorNamespace
 	serviceCAOperatorPodPrefix   = operatorclient.OperatorName
 	serviceCAControllerNamespace = operatorclient.TargetNamespace
 	serviceCAPodPrefix           = api.ServiceCADeploymentName
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -1,0 +1,873 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"slices"
+	"strings"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+
+	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+)
+
+const (
+	agnhostImage = "registry.k8s.io/e2e-test-images/agnhost:2.45"
+)
+
+// Network Policy helpers
+
+func getNetworkPolicy(ctx context.Context, client kubernetes.Interface, namespace, name string) *networkingv1.NetworkPolicy {
+	g.GinkgoHelper()
+	policy, err := client.NetworkingV1().NetworkPolicies(namespace).Get(ctx, name, metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred(), "failed to get NetworkPolicy %s/%s", namespace, name)
+	return policy
+}
+
+func requireDefaultDenyAll(policy *networkingv1.NetworkPolicy) {
+	g.GinkgoHelper()
+	if len(policy.Spec.PodSelector.MatchLabels) != 0 || len(policy.Spec.PodSelector.MatchExpressions) != 0 {
+		g.Fail(fmt.Sprintf("%s/%s: expected empty podSelector", policy.Namespace, policy.Name))
+	}
+
+	policyTypes := sets.New[string]()
+	for _, policyType := range policy.Spec.PolicyTypes {
+		policyTypes.Insert(string(policyType))
+	}
+	if !policyTypes.Has(string(networkingv1.PolicyTypeIngress)) || !policyTypes.Has(string(networkingv1.PolicyTypeEgress)) {
+		g.Fail(fmt.Sprintf("%s/%s: expected both Ingress and Egress policyTypes, got %v", policy.Namespace, policy.Name, policy.Spec.PolicyTypes))
+	}
+}
+
+func requirePodSelectorLabel(policy *networkingv1.NetworkPolicy, key, value string) {
+	g.GinkgoHelper()
+	actual, ok := policy.Spec.PodSelector.MatchLabels[key]
+	if !ok || actual != value {
+		g.Fail(fmt.Sprintf("%s/%s: expected podSelector %s=%s, got %v", policy.Namespace, policy.Name, key, value, policy.Spec.PodSelector.MatchLabels))
+	}
+}
+
+func requireIngressPort(policy *networkingv1.NetworkPolicy, protocol corev1.Protocol, port int32) {
+	g.GinkgoHelper()
+	if !hasPortInIngress(policy.Spec.Ingress, protocol, port) {
+		g.Fail(fmt.Sprintf("%s/%s: expected ingress port %s/%d", policy.Namespace, policy.Name, protocol, port))
+	}
+}
+
+func requireIngressFromNamespace(policy *networkingv1.NetworkPolicy, port int32, namespace string) {
+	g.GinkgoHelper()
+	if !hasIngressFromNamespace(policy.Spec.Ingress, port, namespace) {
+		g.Fail(fmt.Sprintf("%s/%s: expected ingress from namespace %s on port %d", policy.Namespace, policy.Name, namespace, port))
+	}
+}
+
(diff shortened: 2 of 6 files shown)


test/e2e/e2e_test.go
gangwgr, line 37: do we need change in this file?

test/e2e/network_policy.go
```diff
@@ -0,0 +1,467 @@
...
+	g.GinkgoWriter.Printf("mutating NetworkPolicy %s/%s\n", serviceCAOperatorNamespace, defaultDenyPolicyName)
+	mutateAndRestoreNetworkPolicy(ctx, kubeClient, serviceCAOperatorNamespace, defaultDenyPolicyName)
+	g.GinkgoWriter.Printf("mutating NetworkPolicy %s/%s\n", serviceCANamespace, defaultDenyPolicyName)
+	mutateAndRestoreNetworkPolicy(ctx, kubeClient, serviceCANamespace, defaultDenyPolicyName)
+
+	g.By("Checking NetworkPolicy-related events (best-effort)")
+	logNetworkPolicyEvents(ctx, kubeClient, []string{serviceCAOperatorNamespace, serviceCANamespace}, operatorPolicyName)
+}
+
+func getNetworkPolicy(ctx context.Context, client kubernetes.Interface, namespace, name string) *networkingv1.NetworkPolicy {
```
gangwgr: mov helper functions in helpers.go file

test/e2e/e2e_test.go
kaleemsiddiqu: Yes as var serviceCAOperatorNamespace cannot be in two places and keeping it e2e_test.go causing ci tests failures because this var is not accessible in network_policy.go, reason being go compilation limitation of test files.

test/e2e/helpers.go
```diff
@@ -0,0 +1,588 @@
...
+		return false, err
+	}
+	completed, err := kubeClient.CoreV1().Pods(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+	if len(completed.Status.ContainerStatuses) == 0 {
+		return false, fmt.Errorf("no container status recorded for pod %s", name)
+	}
+	exitCode := completed.Status.ContainerStatuses[0].State.Terminated.ExitCode
```
gangwgr: State.Terminated can be nil if the pod failed for non-container reasons (e.g., OOM, eviction). Guard with:

```
if completed.Status.ContainerStatuses[0].State.Terminated == nil {
    return false, fmt.Errorf("container in pod %s not in terminated state", name)
}
```

test/e2e/network_policy.go
```diff
@@ -0,0 +1,122 @@
...
+	defaultDenyPolicyName      = "default-deny-all"
+	operandPolicyName          = "service-ca"
+	operatorPolicyName         = "service-ca-operator"
+)
+
+var _ = g.Describe("[sig-service-ca] service-ca-operator", func() {
+	g.It("[Operator][NetworkPolicy] should ensure service-ca NetworkPolicies are defined", func() {
+		testServiceCANetworkPolicies()
+	})
+	g.It("[Operator][NetworkPolicy] should restore service-ca NetworkPolicies after delete or mutation[Timeout: 30m]", func() {
```
gangwgr: add serial tag here

test/e2e/helpers.go
```diff
@@ -0,0 +1,588 @@
...
+// Network Policy helpers
+
+func getNetworkPolicy(ctx context.Context, client kubernetes.Interface, namespace, name string) *networkingv1.NetworkPolicy {
+	g.GinkgoHelper()
+	policy, err := client.NetworkingV1().NetworkPolicies(namespace).Get(ctx, name, metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred(), "failed to get NetworkPolicy %s/%s", namespace, name)
+	return policy
+}
+
+func requireDefaultDenyAll(policy *networkingv1.NetworkPolicy) {
```
gangwgr, line 41: func requireDefaultDenyAll(policy *networkingv1.NetworkPolicy) {
Checking for empty podSelector and both policyTypes is necessary but not sufficient. A policy with allow rules would still pass. Assert len(policy.Spec.Ingress) == 0 and len(policy.Spec.Egress) == 0.

kaleemsiddiqu:
/retitle [CNTRLPLANE-2894](https://redhat.atlassian.net/browse/CNTRLPLANE-2894):e2e network policy test cases

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
Executed these local tests.

liouk:
This PR should reuse the code that was reviewed and merged in https://github.com/openshift/cluster-authentication-operator/pull/831 as much as possible; at the moment we don't have a common place to put these shared libraries so we've agreed to reuse the reference code from that PR as much as possible. However I see evidence of this code not being used at its latest form (e.g. implementing `joinStrings` instead of using the stdlib `strings.Join()`).

/hold

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,171 @@
...
+	g.It("[NetworkPolicy][Disruptive] should enforce service-ca-operator NetworkPolicies", func() {
+		testServiceCAOperatorNetworkPolicyEnforcement()
+	})
+	g.It("[NetworkPolicy][Disruptive] should enforce cross-namespace ingress traffic", func() {
+		testCrossNamespaceIngressEnforcement()
+	})
+	g.It("[NetworkPolicy][Disruptive] should block unauthorized namespace traffic", func() {
+		testUnauthorizedNamespaceBlocking()
+	})
+})
```
liouk, line 31: These tests don't seem to match any suite defined in `cmd/service-ca-operator-tests-ext/main.go` -- they won't be executed.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,120 @@
...
+const (
+	serviceCANamespace         = operatorclient.TargetNamespace
+	serviceCAOperatorNamespace = operatorclient.OperatorNamespace
+	defaultDenyPolicyName      = "default-deny-all"
+	operandPolicyName          = "service-ca"
+	operatorPolicyName         = "service-ca-operator"
+)
+
+var _ = g.Describe("[sig-service-ca] service-ca-operator", func() {
+	g.It("[Operator][NetworkPolicy] should ensure service-ca NetworkPolicies are defined", func() {
```
liouk: This won't match any suite; maybe we need `[Serial]`?

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,171 @@
...
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+)
+
+var _ = g.Describe("[sig-service-ca] service-ca-operator", func() {
```
liouk, line 15: It looks like this contains only tests that check enforcement only for `openshift-service-ca-operator`, not `openshift-service-ca` -- should we add those as well?

test/e2e/network_policy.go
```diff
@@ -0,0 +1,120 @@
...
+	requireIngressPort(operatorPolicy, corev1.ProtocolTCP, 8443)
+	logIngressFromNamespaceOptional(operatorPolicy, 8443, "openshift-monitoring")
+	requireEgressPort(operatorPolicy, corev1.ProtocolTCP, 5353)
+	requireEgressPort(operatorPolicy, corev1.ProtocolUDP, 5353)
+	requireEgressToNamespace(operatorPolicy, "openshift-dns")
+	logEgressAllowAllTCP(operatorPolicy)
+	logIngressHostNetworkOrAllowAll(operatorPolicy, 8443)
+
+	g.By("Verifying pods are ready in service-ca namespaces")
+	waitForPodsReadyByLabel(ctx, kubeClient, serviceCAOperatorNamespace, "app=service-ca-operator")
```
liouk, line 81: How about service-ca?

```suggestion
	waitForPodsReadyByLabel(ctx, kubeClient, serviceCAOperatorNamespace, "app=service-ca-operator")
	waitForPodsReadyByLabel(ctx, kubeClient, serviceCAOperatorNamespace, "app=service-ca")
```

liouk (changes requested):

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: this gets fetched through cmd/service-ca-operator-tests-ext/dependencymagnet.go

test/e2e/network_policy.go
kaleemsiddiqu: Updated it 
- Added a parallel way to run few tests
- Removed [Distruptive] only based tests and updated tests from network_policy_enforcement.go

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: Included tests for openshift-service-ca as well.

test/e2e/network_policy.go
kaleemsiddiqu: Included this as well.

test/e2e/network_policy.go
liouk: If we're running these in parallel, it looks like there won't be any pod name collisions as they use randomized names; however we got multiple test pods matching the same pod selector, and this makes me wonder whether they could interfere with each other's connectivity assertions? Have you had multiple consistently successful runs of these, or should we just mark them as serial?

kaleemsiddiqu: Yes i have executed multiple successful runs of these in parallel, except the reconciliation one which is serial anyway. I did not see any connectivity assertions issue.

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
local execution is successful.

PR #349, opened 2026-04-28
OCPBUGS-62633: Fix Progressing/Available status conditions on node reboot

tchap:
The main motivation is to have `Progressing=false` when a node is rebooting and not all replicas are available. 

I started with creating unit tests for the current state in the first commit, the following commits actually apply necessary changes. It is more pleasant to review commit-by-commit.

A better long-term solution would be to migrate to the [workload controller](https://github.com/openshift/library-go/blob/master/pkg/operator/apiserver/controller/workload/workload.go) in library-go, which would cause change in operator-level conditions, but we would get all condition management for free.

The overall situation regarding conditions in this operator seems rather arbitrary, but I focused on just fixing the issue and not breaking anything else. 

The following summary is generated by Claude:

## Summary

When a deployment has zero available replicas but is fully up-to-date (all replicas running the current pod template), `syncStatus` was incorrectly reporting `Progressing=True`. This happens during node reboots or pod evictions — the operator is not making progress, the deployment is simply waiting for pods to be rescheduled. Reporting `Progressing=True` in this case can mislead the CVO into thinking an upgrade is still rolling out, delaying or blocking upgrades.

The root cause was a structural bug in `syncStatus`: conditions (`Available=False`, `Progressing=True`) were set inside the deployment loop, but the post-loop code unconditionally overwrote them — every code path after the loop set `Available=True`. This meant `setAvailableFalse` calls inside the loop had no effect.

### Fix

Stop setting conditions inside the loop. Instead, collect flags (`deploymentUnavailableButUpToDate`, `versionUpdatable`) and decide after the loop:

- **Up-to-date but unavailable** (node reboot, pod eviction): `Available=True`, `Progressing=False`, version is set. The operator has nothing to do — it's waiting for infrastructure. We keep `Available=True` to avoid tripping CVO invariant monitors during transient pod rescheduling.
- **Mid-rollout or deleting**: falls through to existing `Progressing=True` paths unchanged.

A `versionUpdatable` guard on the new branch ensures that if one deployment is transiently unavailable but another is mid-rollout, we still report `Progressing=True`.

### Additional changes

- Add `isDeploymentUpToDate` helper to distinguish "unavailable because the node is rebooting" from "unavailable because we're mid-rollout"
- Refactor `isDeploymentStatusComplete` to reuse `isDeploymentUpToDate`
- Migrate `sets.String` to `sets.Set[string]` (deprecated API cleanup)
- Add comprehensive unit tests for `syncStatus` and all deployment status helpers

## Test plan

- [x] `make test-unit` passes
- [X] `make verify` passes
- [ ] e2e: deploy on a cluster with a single worker, drain the worker node, verify `Progressing` stays `False` while the service-ca pod is rescheduling
- [ ] e2e: trigger a deployment rollout (e.g. by changing controller args), verify `Progressing=True` is reported during the rollout and `Progressing=False` after completion

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined deployment status evaluation to better distinguish deployments that are up-to-date but temporarily unavailable, preventing unnecessary "progressing" or version-blocking and improving operator condition accuracy (including setting Available/Progressing appropriately in that case).

* **Tests**
  * Added comprehensive tests covering many deployment rollout scenarios, generation observation, replica edge cases, and status synchronization outcomes.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -96,11 +96,11 @@ func isDeploymentStatusComplete(deploy appsv1.Deployment) bool {
 		deploy.Status.ObservedGeneration >= deploy.Generation
 }
 
-func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA, existingDeployments *appsv1.DeploymentList, targetDeploymentNames sets.String) {
+func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA, existingDeployments *appsv1.DeploymentList, targetDeploymentNames sets.Set[string]) {
 	versionUpdatable := true
 	versionUpdatableAndDeploymentsComplete := true
 	statusMsg := ""
-	existingDeploymentNames := sets.String{}
+	existingDeploymentNames := sets.New[string]()
 	for _, dep := range existingDeployments.Items {
 		existingDeploymentNames.Insert(dep.Name)
 		// If there isn't at least one replica from each deployment, Available=False
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -6,6 +6,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/utils/ptr"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
@@ -86,38 +87,39 @@ func isDeploymentStatusAvailableAndUpdated(deploy appsv1.Deployment) bool {
 }
 
 func isDeploymentStatusComplete(deploy appsv1.Deployment) bool {
-	replicas := int32(1)
-	if deploy.Spec.Replicas != nil {
-		replicas = *(deploy.Spec.Replicas)
-	}
-	return deploy.Status.UpdatedReplicas == replicas &&
-		deploy.Status.Replicas == replicas &&
-		deploy.Status.AvailableReplicas == replicas &&
-		deploy.Status.ObservedGeneration >= deploy.Generation
+	desiredReplicas := ptr.Deref(deploy.Spec.Replicas, 1)
+	return isDeploymentUpToDate(deploy) && deploy.Status.AvailableReplicas == desiredReplicas
+}
+
+func isDeploymentUpToDate(deploy appsv1.Deployment) bool {
+	desiredReplicas := ptr.Deref(deploy.Spec.Replicas, 1)
+	return deploy.Status.ObservedGeneration >= deploy.Generation &&
+		deploy.Status.UpdatedReplicas == desiredReplicas &&
+		deploy.Status.Replicas == desiredReplicas
 }
 
 func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA, existingDeployments *appsv1.DeploymentList, targetDeploymentNames sets.Set[string]) {
 	versionUpdatable := true
 	versionUpdatableAndDeploymentsComplete := true
+	deploymentUnavailableButUpToDate := false
 	statusMsg := ""
 	existingDeploymentNames := sets.New[string]()
 	for _, dep := range existingDeployments.Items {
 		existingDeploymentNames.Insert(dep.Name)
-		// If there isn't at least one replica from each deployment, Available=False
-		reason := "ManagedDeploymentsNotReady"
 		if dep.DeletionTimestamp != nil {
 			statusMsg += fmt.Sprintf("\n%s deleting", dep.Name)
-			setProgressingTrue(operatorConfigCopy, reason, statusMsg)
-			setAvailableFalse(operatorConfigCopy, reason, statusMsg)
 			versionUpdatable = false
 			versionUpdatableAndDeploymentsComplete = false
 			continue
 		}
 		if !isDeploymentStatusAvailable(dep) {
-			statusMsg += fmt.Sprintf("\n%s does not have available replicas", dep.Name)
-			setProgressingTrue(operatorConfigCopy, reason, statusMsg)
-			setAvailableFalse(operatorConfigCopy, reason, statusMsg)
-			versionUpdatable = false
+			if isDeploymentUpToDate(dep) {
+				statusMsg += fmt.Sprintf("\n%s: all replicas are up-to-date but not yet available", dep.Name)
+				deploymentUnavailableButUpToDate = true
+			} else {
+				statusMsg += fmt.Sprintf("\n%s does not have available replicas", dep.Name)
+				versionUpdatable = false
+			}
 			versionUpdatableAndDeploymentsComplete = false
 			continue
 		}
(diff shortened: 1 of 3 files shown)


pkg/operator/status.go
```diff
@@ -86,37 +87,44 @@ func isDeploymentStatusAvailableAndUpdated(deploy appsv1.Deployment) bool {
...
+// isDeploymentUpToDate returns true when the deployment controller has
+// observed the current spec and all replicas use the current pod template.
+func isDeploymentUpToDate(deploy appsv1.Deployment) bool {
+	desiredReplicas := ptr.Deref(deploy.Spec.Replicas, 1)
+	return deploy.Status.ObservedGeneration >= deploy.Generation &&
+		deploy.Status.UpdatedReplicas == desiredReplicas &&
+		deploy.Status.Replicas == desiredReplicas
 }
 
 func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA, existingDeployments *appsv1.DeploymentList, targetDeploymentNames sets.String) {
```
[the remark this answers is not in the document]
tchap: Amended.

pkg/operator/status.go
```diff
@@ -149,6 +157,31 @@ func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA,
...
+		c.setVersion()
+		return
+	}
+	// Deployment is not up-to-date and has no available replicas (e.g. recreate
+	// rollout where old pods are terminated before new ones are ready).
+	if deploymentUnavailableNotUpToDate {
+		reason := "ManagedDeploymentsNotReady"
+		setAvailableFalse(operatorConfigCopy, reason, statusMsg)
+		setProgressingTrue(operatorConfigCopy, reason, statusMsg)
+		return
```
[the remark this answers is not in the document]
tchap: Fixed.

pkg/operator/status_test.go
```diff
@@ -0,0 +1,463 @@
...
+			expectReason:      "ManagedDeploymentsAvailable",
+		},
+		{
+			name:              "available and updated but not all replicas available yet",
+			deployments:       []appsv1.Deployment{makeDeployment("service-ca", 1, 1, 2, 2, 1)},
+			expectProgressing: operatorv1.ConditionTrue,
+			expectAvailable:   operatorv1.ConditionTrue,
+			expectVersionSet:  true,
+			expectReason:      "ManagedDeploymentsAvailableAndUpdated",
+		},
```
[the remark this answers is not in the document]
tchap: Amended.

pkg/operator/status.go
```diff
@@ -86,38 +87,39 @@ func isDeploymentStatusAvailableAndUpdated(deploy appsv1.Deployment) bool {
...
-			setProgressingTrue(operatorConfigCopy, reason, statusMsg)
-			setAvailableFalse(operatorConfigCopy, reason, statusMsg)
-			versionUpdatable = false
+			if isDeploymentUpToDate(dep) {
+				statusMsg += fmt.Sprintf("\n%s: all replicas are up-to-date but not yet available", dep.Name)
+				deploymentUnavailableButUpToDate = true
+			} else {
+				statusMsg += fmt.Sprintf("\n%s does not have available replicas", dep.Name)
+				versionUpdatable = false
+			}
```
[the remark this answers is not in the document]
tchap: This is generally an issue, but we are not gonna fix it now since it breaks e2e tests.

tchap:
/retitle OCPBUGS-62633: Fix Progressing/Available status conditions on node reboot

pkg/operator/status.go
```diff
@@ -86,38 +87,39 @@ func isDeploymentStatusAvailableAndUpdated(deploy appsv1.Deployment) bool {
...
 
-func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA, existingDeployments *appsv1.DeploymentList, targetDeploymentNames sets.String) {
+func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA, existingDeployments *appsv1.DeploymentList, targetDeploymentNames sets.Set[string]) {
 	versionUpdatable := true
 	versionUpdatableAndDeploymentsComplete := true
+	deploymentUnavailableButUpToDate := false
 	statusMsg := ""
-	existingDeploymentNames := sets.String{}
+	existingDeploymentNames := sets.New[string]()
 	for _, dep := range existingDeployments.Items {
```
atiratree, line 107: Why do we track multiple deployments? I can only see one. This is unnecessarily confusing.

pkg/operator/status.go
```diff
@@ -86,38 +87,39 @@ func isDeploymentStatusAvailableAndUpdated(deploy appsv1.Deployment) bool {
...
-		// If there isn't at least one replica from each deployment, Available=False
-		reason := "ManagedDeploymentsNotReady"
 		if dep.DeletionTimestamp != nil {
 			statusMsg += fmt.Sprintf("\n%s deleting", dep.Name)
-			setProgressingTrue(operatorConfigCopy, reason, statusMsg)
-			setAvailableFalse(operatorConfigCopy, reason, statusMsg)
 			versionUpdatable = false
 			versionUpdatableAndDeploymentsComplete = false
 			continue
 		}
```
atiratree, line 114: The status check also seems to expect a very specific deployment with 1 replica.

pkg/operator/status_test.go
```diff
@@ -0,0 +1,463 @@
...
+			deploy: func() appsv1.Deployment {
+				d := makeDeployment("test", 1, 1, 2, 1, 2)
+				d.Spec.Replicas = ptr.To[int32](2)
+				return d
+			}(),
+			expect: false,
+		},
+		{
+			name:   "extra replicas still terminating",
+			deploy: makeDeployment("test", 1, 1, 2, 1, 1),
```
atiratree: FYI: Recreate strategy takes terminating replicas into account. We can even observe the terminating replicas in the status.

pkg/operator/status.go
```diff
@@ -86,38 +87,39 @@ func isDeploymentStatusAvailableAndUpdated(deploy appsv1.Deployment) bool {
...
 		if !isDeploymentStatusAvailable(dep) {
-			statusMsg += fmt.Sprintf("\n%s does not have available replicas", dep.Name)
-			setProgressingTrue(operatorConfigCopy, reason, statusMsg)
-			setAvailableFalse(operatorConfigCopy, reason, statusMsg)
-			versionUpdatable = false
+			if isDeploymentUpToDate(dep) {
```
atiratree, line 116: This could be a false positive. During a workload rollout, we should report progressing until all the replicas become available. Why can't we depend on the Deployment `Progressing` condition instead like we did in https://github.com/openshift/library-go/pull/2128?

pkg/operator/status.go
```diff
@@ -149,6 +151,17 @@ func (c *serviceCAOperator) syncStatus(operatorConfigCopy *operatorv1.ServiceCA,
 		c.setVersion()
 		return
 	}
+	// Deployment is up-to-date but temporarily unavailable (e.g. node reboot,
+	// pod eviction). The versionUpdatable guard ensures this only fires when no
+	// other deployment is mid-rollout or deleting. Keep Available=True to avoid
+	// tripping CVO invariant monitors during transient pod rescheduling.
```
atiratree, line 157: This is incorrect. The deployment is unavailable and since it is a single replica nothing can be done during rescheduling. We either need a monitor exception or increase the number of replicas.

pkg/operator/status_test.go
```diff
@@ -0,0 +1,463 @@
...
+			name: "rolling update in progress - not all pods updated",
+			deploy: func() appsv1.Deployment {
+				d := makeDeployment("test", 1, 1, 2, 1, 2)
+				d.Spec.Replicas = ptr.To[int32](2)
+				return d
+			}(),
+			expect: false,
+		},
+		{
+			name:   "extra replicas still terminating",
```
atiratree, line 217: FYI: Recreate strategy takes terminating replicas into account. We can even observe the terminating replicas in the status.

tchap:
I am going to pause this for now and implement some helpers that can be used for status conditions management as part of https://github.com/openshift/library-go/pull/2128

/hold

